PR #50, opened 2019-08-06
ENH: allow silence warning in Queen_higher

martinfleis:
As libpysal allows silence warnings since 4.1.0, allow them in `Queen_higher` to avoid extensive island warnings.


martinfleis:
Will have to wait until libpysal 4.1.0 is on PyPI and conda-forge. For some reason it was released on GitHub only so far.

PR #88, opened 2019-09-30
DOC: add math formulas, minor changes

martinfleis:


martinfleis:
Added `nose` into dev CI environment as current master of `networkX` would not import without it.

PR #148, opened 2020-02-23
BUG: LineString Z support in nx_to_gdf

### LINKED ISSUE #142: No support for LINESTRING Z `nx_to_gdf` fails
Nice lib! 

I noticed that you wont be able to convert a network back to a gdf if it has elevation data `gdf.has_z == True` . This is due to how `_points_to_gdf` is implemented ->  

```
def _points_to_gdf(net, spatial_weights):
    """
    Generate point gdf from nodes.
    Helper for nx_to_gdf.
    """
    node_xy, node_data = zip(*net.nodes(data=True))
    gdf_nodes = gpd.GeoDataFrame(
        list(node_data), geometry=[Point(i, j) for i, j in node_xy]
    )
    gdf_nodes.crs = net.graph["crs"]
    return gdf_nodes
```

Either some changes has to be made to the code or there should be some mention of this in docs. 
potentially one could add a check  and remove it on the fly.  I had to do this of 3D->2D conversion (`LINESTRING Z` -> `LINESTRING`): 

```
gdf.geometry.apply(lambda geom: shapely.wkb.loads(shapely.wkb.dumps(geom, output_dimension=2)))
```

martinfleis:
Fixes #142 

Make sure Z coordinate is supported in `gdf_to_nx` and `nx_to_gdf`. If LineString Z is passed to `gdf_to_nx`, Point Z is returned for nodes from `nx_to_gdf`.

### DIFF
--- a/momepy/utils.py
+++ b/momepy/utils.py
@@ -222,9 +222,11 @@ def _points_to_gdf(net, spatial_weights):
     Helper for nx_to_gdf.
     """
     node_xy, node_data = zip(*net.nodes(data=True))
-    gdf_nodes = gpd.GeoDataFrame(
-        list(node_data), geometry=[Point(i, j) for i, j in node_xy]
-    )
+    if len(node_xy[0]) == 2:
+        geometry = [Point(i, j) for i, j in node_xy]
+    elif len(node_xy[0]) == 3:
+        geometry = [Point(i, j, k) for i, j, k in node_xy]
+    gdf_nodes = gpd.GeoDataFrame(list(node_data), geometry=geometry)
     gdf_nodes.crs = net.graph["crs"]
     return gdf_nodes
 
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -5,6 +5,8 @@ import networkx
 import numpy as np
 import pytest
 
+from shapely.geometry import LineString
+
 
 class TestUtils:
     def setup_method(self):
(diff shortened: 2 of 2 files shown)


momepy/utils.py
```diff
@@ -222,9 +222,11 @@ def _points_to_gdf(net, spatial_weights):
     Helper for nx_to_gdf.
     """
     node_xy, node_data = zip(*net.nodes(data=True))
-    gdf_nodes = gpd.GeoDataFrame(
-        list(node_data), geometry=[Point(i, j) for i, j in node_xy]
-    )
+    if len(node_xy[0]) == 2:
```
LeonardAukea: Why not use unpacking like `geometry=[Point(*p) for p in node]` ? and maybe a check that the length is at most 3 due to the limitation of the shapely Point object

martinfleis: That's way more elegant than my solution. In that case check is not needed as shapely will raise error anyway. I don't know what else other than error we could do with 4 values here, so it is fine. I'll change this in an upcoming PR adding more compatibility with `osmnx`, it is the same piece of code. Thanks!

LeonardAukea: @martinfleis  Do you have a util for getting the the nearest `Point` on the line? Basically the orthogonal projection? I want to be able to snap points to it's nearest point on a street network and have them added as a nodes in a generated network. I looked  shortly at `pysal spaghetti` but its very slow :/

martinfleis: @LeonardAukea not exactly. I have a function finding nearest line to a point, but it does not project it - https://github.com/martinfleis/momepy/blob/42fc80fc66c1a77035272e30eab04be075495867/momepy/elements.py#L643

Once you have the nearest line you can try to get the point using shapely (https://stackoverflow.com/a/24440122/9978945), but I am not sure if it will be faster than spaghetti. I want to do it at one point as it allows measuring reach from points along the network, but it likely won't happen soon.

PR #182, opened 2020-06-25
ENH: Fixed squareness for non-Polygon geom types

AleFeli:
I added a condition for multi-polygon geometries


martinfleis:
Thanks :). The failure in Travis CI is a known regression in the latest GeoPandas, so we can ignore that and merge this fix. There might be more of this kind, so feel free to do the same fix for those if you find them. Returning `np.nan` makes more sense than raising an error here.

PR #229, opened 2020-11-26
DOC: plot graphs

martinfleis:
Plot primal and dual graphs into geographic space.


martinfleis:
Failing tests are due to https://github.com/networkx/networkx/pull/4392 hence fix in tests.

PR #261, opened 2021-02-09
EHN: Add SkyViewFactor

jonwangio:
sky view factor initiated.


martinfleis:
@jonwangio Can you add BVD as a separate PR from a separate branch?

martinfleis:
Given this hasn't happened in the last 3 years, I am not sure it will :). Since the result is a grid, it is also a bit different approach than the rest of momepy. I will close this now to clear the PR backlog but if there's an interest from someone, we can always revisit.

PR #262, opened 2021-02-11
TST: fix tests precision

martinfleis:


martinfleis:
Remaining failure needs a bit more exploration.

PR #264, opened 2021-02-12
ENH: Simple Building Volume Density

jonwangio:
Add BVD

### DIFF
--- a/momepy/bvd.py
+++ b/momepy/bvd.py
@@ -1,0 +1,110 @@
+#-------------------------------------------------------------------------------
+# Program Name: Simple Building Volume Density (BVD) for building morphological
+# Purpose: for urban morphological metrics of MOMEPY
+
+# This Python 3 programs depends on: math, shapely, fiona and geopandas
+
+# Version:     0.1 
+#              Functionalities:
+#              1. Building volume density
+#              2. Volume density computed as the ratio of 
+#              building volumes to a virtual box defined by the highest building in an unit area...
+
+# Author:      Momepy
+#
+# Created:     04/02/2021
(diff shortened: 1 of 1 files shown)


martinfleis:
@jonwangio Are you planning to turn this into a momepy class or do you want me to do that based on your code? See for example implementation of `SharedWallsRatio` - https://github.com/martinfleis/momepy/blob/be19ab9fb0a4daa40c82f6ac89c3920ac632df2b/momepy/distribution.py#L110

I have no problem doing that but it may take some time. 

Normally, to add a new function it should be a class with a similar signature as the existing ones, with docstring and tests. I can guide you through the implementation if needed.

Just want to know what to expect with this and the other PR.

jonwangio:
> 
> 
> @jonwangio Are you planning to turn this into a momepy class or do you want me to do that based on your code? See for example implementation of `SharedWallsRatio` -
> 
> https://github.com/martinfleis/momepy/blob/be19ab9fb0a4daa40c82f6ac89c3920ac632df2b/momepy/distribution.py#L110
> 
> I have no problem doing that but it may take some time.
> 
> Normally, to add a new function it should be a class with a similar signature as the existing ones, with docstring and tests. I can guide you through the implementation if needed.
> 
> Just want to know what to expect with this and the other PR.

@martinfleis Would be great if you can turn them into what they should be for this time, and I can compare and try to mimic next time. Thanks.

martinfleis:
I think that this can be done using one or two calls to `Graph.describe()` these days. See http://docs.momepy.org/en/latest/user_guide/weights/examples.html#Neighborhood-description

PR #276, opened 2021-05-28
ENH: continuity in street networks COINS

amorfinv:
Adapted COINS algorithm for momepy. It is not yet a function but the idea would be that it has the following structure:

Argument:
- edge geodataframe(gdf)

Returns:

- new gdf with pre_merge values
- new gdf with stroke groups
- original gdf with new attribute containing stroke groups.


amorfinv:
I adapted COINS so can be accessed via momepy. See below for a nice example to run.

```python
    import osmnx as ox
    import momepy

    streets_graph = ox.graph_from_place('Pisek, Czechia', network_type='drive')
    streets_graph = ox.projection.project_graph(streets_graph)

    streets = ox.graph_to_gdfs(ox.get_undirected(streets_graph), nodes=False, edges=True,
                                    node_geometry=False, fill_edge_geometry=True)

    stroke_group = momepy.COINS(streets)

    f, ax = plt.subplots(figsize=(10, 10))
    stroke_group.plot(ax=ax, column='stroke_group')
    ax.set_axis_off()
    plt.show()
```
strokes

martinfleis (commented):
Thanks a lot, this looks really good.

I would maybe change the API to class-like (see the comment in code).

Apart from that, we will need some tests. Do you know how to do them? I can do some if needed. 

Once we have tests, I would merge it and then make a series of smaller PRs refactoring individual parts to make it a bit faster. It is quite slow at the moment.

Do you know how slower/faster are individual parts compared to the original command-line implementation?

momepy/coins.py
```diff
@@ -0,0 +1,572 @@
...
+    def splitLines(self):
+        outLine = []
+        tempLine = []
+        self.tempArray = []
+        n = 0
+        #Iterate through the lines and split the edges
+        idx = 0
+        for line in self.lines:
+            for part in _listToPairs(line):
+                outLine.append([part, _computeOrientation(part), list(), list(), list(), list(), list(), list(), self.uv_index[idx]])
```
martinfleis, line 143: What are these empty lists for?

amorfinv: These lists are used in the negotiation to choose the stroke group.  If you look at `create_gdf_premerge()` you can see what each list represents.

- linksP1
- linksP2
- bestP1
- bestP2
- P1Final
- P2Final

These empty lists get filled in the `getLinks()`, `bestLink()` and `crossCheckLinks()` methods.

You can refer to Table 2 from the [paper](https://journals.sagepub.com/doi/10.1177/2399808320967680) to see what each variable represents.

If you compare to the original [code](https://github.com/PratyushTripathy/COINS/blob/master/PythonTool/200330_NetworkContinuityMultiprocessing_PT_Rework_Final.py) (line 277), you will see that I added the last item of outLine. This is so that I can track the original edge that each segment belongs to.

momepy/coins.py
```diff
@@ -0,0 +1,572 @@
...
+"""
+
+import sys, math, time, multiprocessing
+from functools import partial
+import numpy as np
+from shapely.geometry import Point, LineString, MultiLineString
+from shapely import ops
+import geopandas as gpd
+
+def COINS(edge_gdf, angle_threshold=0, out_gdf='stroke_attribute'):
```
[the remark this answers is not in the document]
amorfinv: Ok I will make the change to a class. Shall I change the name of `line` to `COINS`?

amorfinv:
I am not sure how to run the tests, but I am happy to learn if you don't have time.

Regarding the performance, the [example](https://github.com/martinfleis/momepy/pull/276#issuecomment-850821780) above gives me the following times

- `getLinks()` 2.63 seconds
- `mergeLines()` 2.64 seconds
- total: 5.39 seconds

In comparison, the command line implementation using a shapefile of the same street graph gives:

- `getLinks()` 0.53 seconds
- `mergeLines()` 0.41 seconds
- total: 1.16 seconds

`getLinks()` and `mergeLines()` are the only methods using a multiprocessing pool

amorfinv:
Ah I see, makes sense. I will make the change now.

momepy/coins.py
```diff
@@ -0,0 +1,539 @@
...
+
+        self.split = outLine
+
+    def uniqueID(self):
+    #Loop through split lines, assign unique ID and
+    #store inside a list along with the connectivity dictionary
+        self.unique = dict(enumerate(self.split))
+
+    def getLinks(self):
+        global result
```
martinfleis, line 131: This is something we don't need within a class. This line can go and all `results` can become `self.results`.

momepy/coins.py
```diff
@@ -0,0 +1,539 @@
...
+                self.unique[edge][4] = 'DeadEnd'
+                
+            if len(p2AngleSet)!=0:
+                val2, idx2 = max((val, idx) for (idx, val) in enumerate(p2AngleSet))
+                self.unique[edge][5] = self.unique[edge][3][idx2], val2
+            else:
+                self.unique[edge][5] = 'DeadEnd'
+
+    def crossCheckLinks(self, angleThreshold):
+        global edge, bestP1, bestP2
```
martinfleis, line 189: Same here. No need for global variables.

momepy/coins.py
```diff
@@ -0,0 +1,539 @@
...
+        # create empty list that will contain attributes of stroke
+        stroke_group = list()
+
+        for edge in self.uv_index:
+            stroke_group.append(inv_edges[edge])
+        
+        return pd.Series(stroke_group, index=self.edge_gdf.index)
+
+#Set recurrsion depth limit to avoid error at a later stage
+sys.setrecursionlimit(10000)
```
martinfleis, line 339: I know this comes from the original code but do you know where does it cause an issue? I'd rather not have this here if we can avoid it.

martinfleis (commented):
Thanks! Few minor comments in the code.

After that, I'll add some super simple tests and merge. 

This is a complex code so it is pretty hard to review unless you have a day for that (which I don't atm). 

I am certain that some of the performance hits we see can be avoided and we may even be faster than the original code in the end. However, I'll need to find some time to actually look into details trying to understand the internal logic to figure out how to refactor it. I'll try to do it before we'll release 0.5.0 but my guess is that it won't happen sooner than in July.

No matter the performance, it is amazing that we have a functional version of the COINS algorithm here, so thanks a lot to all involved! If Niki Patrinopoulou and Ioannis Daramouskas want recognition as contributors on GitHub, we can figure out some PRs for that.

momepy/coins.py
martinfleis: @PratyushTripathy do you remember the reason for this?

momepy/coins.py
amorfinv: Sounds good. From looking at the code, it seems like I don't need to make them `bestP1` or `bestP2` class variables. I am a but unsure about `edge`. However, it seems like if I just comment out this line, things are still working.

martinfleis:
> Just will wait on what we decide about the recursion limit before updating the PR.

Remove it, I don't want to change the system-wide recursion limit here. If it is going to cause trouble 1) user can change it themselves, 2) we can fix that :).

amorfinv:
Hopefully, everything is good for an initial implementation. 

There were three other PhD students (Niki, Ioanis, and Andrei) that helped me figure out the change from shapefile to gdf. If it is not too much of a burden, it would be cool if they were also contributors. We got to impress our supervisors haha.

Their handles are: @andubadea, @nipatrin, and @idaramouskas

Best,
Andres

PratyushTripathy:
Thank you @amorfinv, @martinfleis, @andubadea, @nipatrin, and @idaramouskas!

This looks very good. Sorry for not being able to actively participate in this. Will look forward to its release.

@amorfinv, can the new attributes containing stroke groups be in the pre-merge gdf itself so that there are only two outputs? Otherwise, the pre-merge gdf will be same as the original gdf, except for the attributes (please correct me if I am wrong). It Will be useful to have fewer variables stored in the memory, especially when a user is processing large street network on a low memory machine.

amorfinv:
Hello @PratyushTripathy,

It is a great script and has already helped us quite a bit. We are currently using it to help us build urban airspaces. Check out this video that we made with our air traffic simulator ([BlueSky](https://github.com/TUDelft-CNS-ATM/bluesky)).

https://user-images.githubusercontent.com/78442543/121365961-5fa68700-c939-11eb-8f0a-ae915ab4ec4d.mov

We are tracking the stroke group of the drone. Our hope is to use this information to make safe urban airspace.

-------------------------------

Regarding your question,

It is a little bit different. The following outputs should be what came out of your original code:

```python
premerge = coins.premerge() 
stroke_gdf = coins.stroke_gdf()  
```

So `premerge` contains all the line segments and is defined as follows:

```python
premerge = gpd.GeoDataFrame(myList, columns=['UniqueID','Orientation','linksP1','linksP2','bestP1','bestP2','P1Final','P2Final', 'geometry'], crs=gdfProjection)
```

I guess we could also add a stroke_group attribute to `premerge`. Maybe it is something to consider

`stroke_gdf` combines all the segments of each group and the gdf is defined this way,

```python
stroke_gdf = gpd.GeoDataFrame(myList, columns=['stroke_group', 'nSegments', 'geometry'], crs=gdfProjection)
```

We added the third output (a pandas series), 

```python
stroke_attr = coins.stroke_attribute()
```

This was easier to work with for our case. It just adds an attribute to the gdf given to create a `coins` object. This would be the `streets` gdf seen below:

```python
coins = momepy.COINS(streets)
```

In our case, the `streets` gdf comes from `osmnx`. I am not sure how familiar you are with `osmnx`. But basically, each edge can contain several line segments. This would mean that the `premerge` and `streets` gdf are of different size. 

I think there is an option where you can actually make it so that `osmnx` makes each line segment a unique edge. However, this is not something useful for us.

To summarize
- `streets` is a gdf from `osmnx`. In the [test example](https://github.com/martinfleis/momepy/pull/276#issuecomment-850821780) it has 778 rows.
- `premerge` is a gdf with individual line segments as edges. Example has 3958 rows.
- `stroke_gdf` is a gdf where each edge is a stroke group. Example has 254 rows.
-  `stroke_attr` is a pandas series of length 778 that tells you what stroke group the edges from `streets` belong to.

Best,
Andres

PratyushTripathy:
Hi @amorfinv,

Thanks for your prompt and detailed response. Happy to hear about your project, it looks really cool! 

I totally agree with you on the above points. The only thing is that I was under the impression that osmnx returns street network as graph where each edge is a standalone unit, which made me assume that the original street gdf would be similar to the premerge gdf (exploded lines). But like you mentioned, that can vary depending on how the user pulls data from osmnx.

Regards,
Pratyush

momepy/coins.py
PratyushTripathy: @martinfleis there is a recursion function that reaches maximum depth when the street network is large. While processing it for bigger cities like Delhi, the original code hits the recursion limit, so I had to increase it to be able to process it.

martinfleis: @PratyushTripathy Can point to the specific bit in the code where this happens? Just to make a not there.

PratyushTripathy: @martinfleis Not able to find it. It's been a while, I don't exactly remember the function. It could be that I re-wrote the function and solved the issue but forgot to remove this line. But to be sure, I'm running a large city again using the old code without increasing the recursion limit. Will get back on this.

PratyushTripathy: @martinfleis I processed a large city twice, the first time with that line (recursion limit increase) commented and the second time by setting the limit to 1000 (I guess that's Python's default). It processed easily. Looks like this was something for an older version of code that I missed removing. Please go ahead and remove it. And thanks for flagging :)

PratyushTripathy:
@amorfinv One small thing, in the final output (strokes), can we add a piece of code that computes the length of each stroke and export it in the attributes table? Would be very helpful to open the SHP in a GIS environment and classify directly using length. Currently, this is an additional manual step after COINS generates the final output.

momepy/coins.py
martinfleis: @PratyushTripathy Brilliant! Thanks for testing it out!

martinfleis:
> Currently, this is an additional manual step after COINS generates the final output.

If I may, I would keep it like that. Measuring length is a simple `.length` call if you need it and we should keep the function lean if possible.

amorfinv:
@PratyushTripathy 

I am also interested in getting stroke_lengths for my case since it can give some sort of street priority. It will help me divide up traffic.

I was doing it this way.

```python
# get stroke lengths
stroke_lengths = stroke_group_gdf.length

# create empty list that will contain attributes of stroke length
stroke_length_attrib = list()

# loop through dataframe
for row in streets.iterrows():
    stroke_group = row[1]['stroke_group']
    stroke_length_attrib.append(stroke_lengths[stroke_group])

# add attribute to streets gdf
streets['stroke_length'] = stroke_length_attrib
```

Essentially, I do what @martinfleis before exporting it to `.gpkg` or `.shp`. I take the second output of COINS, do a `.length` and then match it to the `streets` gdf.

I think there is probably a smarter way by using some pandas dataframe method. Maybe `df.merge()` or `df.join()`?

It seems to work nicely in QGIS.

Screen Shot 2021-06-10 at 17 47 54

The thick red line are longest stroke lengths.

amorfinv:
I wanted to suggest a future improvement. I am still not sure if this improvement should be part of `COINS` or is something that should be done before giving the network to `COINS`.

I noticed that sometimes choosing the street links with the largest interior angle is not always the ideal choice. See for example this exit near a highway.

Screen Shot 2021-06-10 at 17 59 07

If you look at the white circle you will see that stroke number 5 has two choices to continue. I guess because of the way the data is provided, exiting the highway is the best choice if you look at the interior angle. However, I think it makes more sense to actually continue where group 32 starts.

Perhaps in cases where the choice is very close, there can be another check based on the `highway` tag from OpenStreetMap that is also available with an `osmnx` attribute. In this case the edge in stroke 5 that is part of the highways is labeled as `highway=trunk`. However, the edge in the exit that is also part of stroke 5 is labeled as `highway=trunk_link`. Lastly, the starting edge of group 32 is labeled as `highway=trunk`. Here is the [info](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dtrunk_link) from OSM. 

I am unsure if this should be part of `COINS` or if it should be fixed before. At first I thought that this could be solved with `osmnx.simplification.consolidate_intersections()` But it did not work for me. Maybe there is another way to do this before `COINS`.

Also, in cases where it is not a highway, the OSM tag won't really work. So maybe in cases where the choice is not clear, it can be based on street names?

What do you think @PratyushTripathy @martinfleis?

martinfleis:
@amorfinv this is a topologically correct network, there's nothing to be fixed before it goes to the COINS algorithm.

I can imagine having an additional check inside `COINS` which would compare two potential directions and if their difference is under a certain threshold it would prefer the one with the same attribute. In any case, that should certainly be done later as an independent PR (which is very welcome!).

PratyushTripathy:
@amorfinv I completely agree with you. This is something that has been on the list since the beginning. Roundabouts are also a problem. But there are two problems, first is that street network data can come from sources other than OSM too. I have come across instances where OSM data for a city (mostly in developing countries) is not that great and alternate sources (resource intensive) are preferred. In that case, the attributes are not present. So I did not want to restrict COINS to OSM data.

Second, in many cities of the Global South, clear street naming conventions are not followed. As a result, street network data with clear street labelling is scarce and name-based street hierarchy would not work very well. For example, consider a continuous street located in Bengaluru, India at 77.58404 east and 13.00236 west, shown in the figure below. The continuous street segment has names ‘Bellary Rd’, ‘Sankey Rd’ and ‘Guttahalli Main Rd’ in the Google Maps data, a name-based hierarchy will not identify these as segments of a continuous street. In addition, the same street in the OpenStreetMap data has names ‘T Chowdiah Road’ and ‘Ramana Maharshi Road’.
image

Google Maps link
https://www.google.co.in/maps/@13.0031311,77.5840167,17.54z

Open Street Map link
https://www.openstreetmap.org/#map=18/13.00273/77.58403

Will have to really think to find a global solution. But yeah, as @martinfleis mentioned, we can do this as an independent PR later, perhaps an optional function in the function so that non-OSM data can also be processed.

Thank you both!

martinfleis:
> I did not want to restrict COINS to OSM data

Yeah, fully agree with that. The function @amorfinv mentions can be based on an arbitrary attribute no matter if it comes from OSM or other sources. And yes, fully optional because you won't have it in a lot of cases.

amorfinv:
Sounds great. It makes sense to make it optional. 

@PratyushTripathy, I haven't tried it bur perhaps following the methodology from `osmnx.simplification.consolidate_intersections()` can help deal with roundabouts.

martinfleis:
Hi @amorfinv,

I have pushed some changes. To sum up - I have removed multiprocessing and reimplemented `getLinks`. 

It is now way faster than it was and there's still space for improvements. However, I'll probably skip those for now as it is currently fast enough.

On the sample linked above, I now get `360 ms` to do

```py
coins = momepy.COINS(streets)
stroke_gdf = coins.stroke_gdf()
```

instead of something like 5s before. The original script takes 1.92 seconds on my machine.

On a larger network (whole Prague), the same bit takes 4.5 seconds, compared to 38 seconds using the original script by @PratyushTripathy. 

Decent speedups I think :). I just hope I haven't made some error but the output of the new `getLinks` should be exactly the same as the previous one.

I'll try to push tests tomorrow and merge it. Can you try to test the changes in the meantime?

amorfinv:
@martinfleis Wow that is great!

On the example it now takes me 230 ms. super fast! It seems to be working fine.

I also tested in a neighborhood in Vienna and it is also correctly getting the stroke groups. It went from 6.5s to 35s. 

-Andres

PratyushTripathy:
This is awesome! I didn't see this coming--I expected some speedups but not to this extent. Quite curious to run it for large cities and compare with the original script.

@martinfleis I really like when you say "there's still space for improvements" ;)

martinfleis:
@amorfinv what is the use case for `premerge()` dataframe? When is it useful in its current form?

amorfinv:
Perhaps @PratyushTripathy will have a better idea. From looking at the publication it was used to see how the decisions get made before merging. So at an intersection you can see interior angles between all edges and then which edge was selected for which group, etc..

I can't really think of a use case, other than  it will be useful when making improvements to the algorithm.

PratyushTripathy:
@amorfinv You are right. There is no use case as such. I kept it so that there is something to cross-check from in case of doubt on the algorithm.

@martinfleis If removing `premerge()` dataframe improves the execution time or memory load, etc., please take a call. Otherwise it can be kept as a class variable maybe?

martinfleis:
It is not called anywhere unless you explicitly want it, so it makes no difference performance-wise. I'll turn it into a private function so it is not a part of the public API to make that cleaner.

martinfleis:
Thanks a lot to everyone involved!

@amorfinv if you want to add that parameter giving a preference to an attribute over angle, let me know. Otherwise, I'll cut the release in the following days.

amorfinv:
Awesome! It was fun.

I am thinking about doing this functionality for my project so I will keep you updated. However, I don't think I will be able to get to it until August.

-Andres

PR #277, opened 2021-07-09
A few typos / wording changes

matthew-law:
Let me know if there are naming conventions for commits that I've missed somewhere


martinfleis:
Thanks! 

> Let me know if there are naming conventions for commits that I've missed somewhere

Nothing serious, no.

PR #280, opened 2021-07-19
ENH: add dropna keyword to `Unique`

### LINKED ISSUE #279: momepy.Unique counts Nans while it shouldn't
https://github.com/martinfleis/momepy/blob/896cb70ef80bfca07fd12dd5f3da827613367ab6/momepy/diversity.py#L778
This line should be changed to exclude nans in the computation of unique elements. Possible solution: add a keyword `include_nan` and if False switch this line to values_list.dropna().unique()

martinfleis:
Closes #279


martinfleis:
The failure of the docs is fixed in geopandas master. Merging.

PR #285, opened 2021-07-28
Update two broken URLs

matthew-law:
I haven't been able to access https://guide.momepy.org/ either, which is also still in the readme, but I'm not 100% sure the error isn't on my end so haven't changed this yet

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Momepy aims to provide a wide range of tools for a systematic and exhaustive ana
 Comments, suggestions, feedback, and contributions, as well as bug reports, are very welcome.
 
 ## Getting Started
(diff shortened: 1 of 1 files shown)


martinfleis:
guide.momepy.org doesn't exist anymore. It has been part of the main docs for a while. You found it somewhere else as well? Can you also change it?

matthew-law:
> guide.momepy.org doesn't exist anymore. It has been part of the main docs for a while. You found it somewhere else as well? Can you also change it?

Should be done now c85f4ac79a41da7cbab52aaf49ce3a1db7f7a14b

martinfleis:
Thanks!

PR #288, opened 2021-08-02
BUG: properly clip enclosures by limit

### LINKED ISSUE #270: Enclosure has problems with rivers
Although I specified streets and land as spatial constraints for generating the enclosures, the momepy.enclosures wrongly creates enclosures in rivers. See the yellow areas in the pic. 
wards

### LINKED ISSUE #287: momepy.enclosures doesn't appear to observe limit parameter
When building enclosures with `momepy.enclosures`, the `limit` parameter doesn't seem to be observed. Looking through the code, my best guess is that [this line](https://github.com/martinfleis/momepy/blob/v0.4.4/momepy/elements.py#L988) of the source adds the `primary_barriers` and the `limit` together when defining the barriers to create the enclosures within, thereby expanding the supplied `limit` to also include the area bounded by the `primary_barriers`, but I could be wrong.

[Here](https://gist.github.com/matthew-law/7b5c29ae31a72f02c43f195d37b26705) is a notebook showing an example – the final plot shows enclosures outside of the supplied `limit`.

Alternatively, I've misunderstood the docs and `limit` is not meant to be the definitive geographical boundary within which the enclosures are generated.

martinfleis:
Ensures that generated enclosures are fully within a `limit` (when provided) by calling `geopandas.clip`.

@matthew-law can you test this patch on your data? I'd also like to know what is the performance hit of doing this (it may be significant). I may add a keyword to controlm whether we should try to clip or not.

closes #287, closes #270

### DIFF
--- a/momepy/elements.py
+++ b/momepy/elements.py
@@ -944,7 +944,11 @@ def get_network_ratio(df, edges, initial_buffer=500):
 
 
 def enclosures(
-    primary_barriers, limit=None, additional_barriers=None, enclosure_id="eID"
+    primary_barriers,
+    limit=None,
+    additional_barriers=None,
+    enclosure_id="eID",
+    clip=False,
 ):
     """
     Generate enclosures based on passed barriers.
@@ -969,6 +973,8 @@ def enclosures(
         (Multi)LineString geometry is expected.
     enclosure_id : str (default 'eID')
         name of the enclosure_id (to be created).
+    clip : bool (default False)
+        if True, the enclosures will be clipped to the extent of the limit (if given)
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


matthew-law:
I've set up a development environment + build and tested with [this example](https://gist.github.com/matthew-law/7b5c29ae31a72f02c43f195d37b26705), but it doesn't have seemed to fix the issue:

image

It may be that I'm doing something wrong on my end, but FWIW I didn't change any of the code from before, just ran it with the updated version.

martinfleis:
You need to pass `clip=True`.

matthew-law:
Oops – fixed that and it works as desired!

It definitely slows down enclosure creation (which is already one of the bottlenecks in terms of speed):

Existing version:
image

With clip:
image

If there are any more scientific/rigorous tests you'd like me to run let me know – these are obviously just a small example but probably indicative of a general difference in speed.

martinfleis:
@matthew-law I have changed the implementation here and there's no visible performance hit on `clip=True` anymore. I'll keep it False by default as you may pass LineStrings as a limit in which case this _clipping_ would not case as it checks if tessellation cells are within the limit.

PR #296, opened 2021-08-23
PERF: use dask.bag in Tessellation

### LINKED ISSUE #295: `momepy.Tessellation` doesn't appear to run faster on a machine with many more cores
On my quad-core MacBook:
image

The same code on a 16+ core machine (the GDSL box):
image

(`use_dask` is made explicit in the second case but should be the same for both by default)

martinfleis:
Fixes #295 

@matthew-law You were right, in my test, the single-core implementation was actually a bit faster than parallel. This seems to fix the problem. 

I am a bit worried though. I used `bag` initially and then when moving the code to momepy, I have switched to dask.Series. I just don't remember why. Can you try this version on your data? My assumption is that it should go from 30 minutes to about 5.

### DIFF
--- a/momepy/elements.py
+++ b/momepy/elements.py
@@ -454,7 +454,7 @@ class Tessellation:
 
         if use_dask:
             try:
-                import dask.dataframe as dd
+                import dask.bag as db
                 from dask.system import cpu_count
             except ImportError:
                 use_dask = False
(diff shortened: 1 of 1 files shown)


matthew-law:
@martinfleis even better than predicted, thanks for having a look and fixing the issue!

image

PR #302, opened 2021-10-01
ENH: Add SNDi

amorfinv:
Add SNDi to momepy

### DIFF
--- a/momepy/__init__.py
+++ b/momepy/__init__.py
@@ -11,6 +11,7 @@ from .utils import *
 from .weights import *
 from .preprocessing import *
 from .coins import *
+from .sndi import *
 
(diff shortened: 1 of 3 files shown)


amorfinv:
Thanks @martinfleis

I calculated a couple of SNDi's for some cities and then compared them to the ones from the paper. However, they are quite off. I think that means one of two things:

1. I am calculating the metrics wrong
2. I am incorrectly pre-processing the graph.

I guess it could also be both. I think next step is to ask the authors a bit more about the pre-processing so I will do that. I could also go through their code and see.

-Andres

amorfinv:
It has been quite some time and I am not sure if I will be able to work on this in the near future. 

In the meantime I will close this PR.

-Andres

Issue #308, opened 2021-10-03
API: move from object-based design to function-based


martinfleis:
I want to propose a big API overhaul, moving away from the current object-based API to a simpler function-based one.

### Current situation

At the moment, the functionality can be split into several groups:

1. basic characters with a single output

The majority of characters takes input variables and their main output is a single pandas Series. The usual pattern looks like this (note the `.series`):

```py
gdf['volume'] = momepy.Volume(gdf, heights='height_col').series 
```

2. characters with multiple outputs

Some characters result in multiple Series'. Those were initially the main reason to implement object-based API.

```py
street_prof = momepy.StreetProfile(streets_df, buildings_df, heights='height')
streets_df['width'] = street_prof.w
streets_df['deviations'] = street_prof.wd
```

3. characters measured on networkx Graph

Characters measured by the `graph` module use functions. They take a network, measure stuff and save the result as a node or edge attribute of a given network and return its copy with that attribute.

```py
network_graph = mm.betweenness_centrality(network_graph)
```

4. classes generating morphological features

Classes that are used to generate new geometries (like `Tessellation`) return a new GeoDataFrame but also Series' with IDs in some cases to link the resulting gdf to the original one.

```py
blocks = mm.Blocks(tessellation_df, streets_df, buildings_df, 'bID', 'uID')
blocks_gdf = blocks.blocks
buildings_id = blocks.buildings_id
```

5. additional utilities that are usually functions

```py
buildings_df['nID'] = momepy.get_network_id(buildings_df, streets_df, 'nID')
```

### Proposal

I would like to simplify all that and turn existing classes into functions returning numpy arrays. There are a few reasons for that:

- calling the `.series` attribute is annoying and there's almost no value in other attributes we currently provide
- `dask_geopandas` compatibility - while you can use existing classes within `map_partitions`, it is not straightforward, especially for classes with multiple outputs.
- lighter code - wrapping everything in classes generates a lot of unnecessary code. The switch should in the long-term allow easier maintenance.
- compatibility with the rest of PySAL, especially `esda`

So, the proposed changes to the groups listed above:

1. basic characters with a single output

A simple function that returns a numpy array.

```py
gdf['volume'] = momepy.volume(gdf, heights='height_col')
```

2. characters with multiple outputs

A function that returns a tuple of arrays (for example like `gdf.sindex.query_bulk` does)

```py
width, deviations = momepy.street_profile(streets_df, buildings_df, heights='height')
streets_df['width'] = width
[...]

PR #330, opened 2022-01-05
BUG: fix FormFactor

### LINKED ISSUE #329: BUG: FormFactor uses incorrect formula
I just realised that `FormFactor` is using an incorrect formula. 

The formula in the original paper by Bourdic et al says `S/(V^2/3)` where `S` is surface. However, I have implemented it as `A/(V^2/3)`, where `A` is an area of a polygon, not the theoretical surface of the whole building.

I think we should fix this with `S` defined as `perimeter * height + area`.

I'll try to get it done today and cut the 0.6.0 release to have it ready for the PySAL meta release later this month.

martinfleis:
Closes #329

This reminded me how much I actually hate myself for the design of these classes and the need for #308.

### DIFF
--- a/momepy/shape.py
+++ b/momepy/shape.py
@@ -32,12 +32,34 @@ __all__ = [
 ]
 
 
+def _form_factor(height, geometry=None, area=None, perimeter=None, volume=None):
+    """helper for FormFactor"""
+    if area is None:
+        area = geometry.area
+    if perimeter is None:
+        perimeter = geometry.length
+    if volume is None:
+        volume = area * height
+
+    surface = (perimeter * height) + area
+    zeros = volume == 0
+    res = np.empty(len(geometry))
+    res[zeros] = np.nan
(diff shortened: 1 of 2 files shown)


martinfleis:
There will be a CI failure in the dev env that can be ignored. It is coming from regressions in geopandas main.

momepy/shape.py
```diff
@@ -32,12 +32,34 @@
 ]
 
 
+def _form_factor(height, geometry=None, area=None, perimeter=None, volume=None):
```
jGaboardi, line 35: Could `geometry` actually be `None` here or should it be a positional argument?

martinfleis: Ah, not at the moment as I use it on line 46. But in theory yes, if you pass all the other values as precomputed arrays. I'll keep it positional for now and revisit during the refactoring.

jGaboardi (approved):

PR #332, opened 2022-01-07
MAINT: replace LooseVersion with packaging.version

martinfleis:
`LooseVersion` is deprecated. The warning recommends replacing it with `packaging.version`, which is exactly what I am doing here. Apart from that, I have also ran `isort` and temporarily turned off `fail-fast` in CI, while there's that GeoPandas regression.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -25,6 +25,7 @@ jobs:
       run:
         shell: bash -l {0}
     strategy:
+      fail-fast: false
       matrix:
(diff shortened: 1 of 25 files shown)


martinfleis:
@jGaboardi @sjsrey this is likely something we'll have to across PySAL.

jGaboardi:
Should `packaging` also be added to `requirements.txt`?

jGaboardi (approved):

jGaboardi:
Good to merge now I think.

PR #335, opened 2022-01-10
Add issue templates

martinfleis:
I'd like to experiment with issue forms here, to ensure we get the proper info we need. 

Templates are _borrowed_ from  the amazing [`executablebooks`](https://github.com/executablebooks/.github/tree/master/.github/ISSUE_TEMPLATE) project and adapted where needed. They actually deployed these forms across the whole org via `.github` repository (see the link). I didn't know that is possible but it could likely make our lives easier as it allows making the same changes in all PySAL repos at once. 

cc @sjsrey @jGaboardi @knaaptime


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,80 @@
...
+        Provide a short description (one or two sentences) about the problem. What did you expect to happen, and what is actually happening?
+
+        If possible, provide full tracebacks (error messages) that you've encountered.
+      placeholder: |
+        The behavior of XXX doesn't work as expected. When I do YYY, I expected ZZZ to occur, but instead WWW happens.
+
+        Here's an error message I ran into...
+
+        ```python
+        momepy.Tessellation(df)
```
jGaboardi, line 19: The code blocks don't seem to be rendering properly here (or maybe they are?). Is this desired? See [here](https://github.com/pysal/momepy/blob/275ab12b464d4eb17273e7ed1d9e93f393faac6e/.github/ISSUE_TEMPLATE/bug_report.yml).

jGaboardi:
@martinfleis Very interesting...

knaaptime:
nice! im definitely onboard

.github/ISSUE_TEMPLATE/bug_report.yml
martinfleis: Can you send a screenshot? On my side it doesn't fit the text area but that is different when you click on it in a live form.

jGaboardi: I'm wondering if the code snips inside the entry cells should/could be rendered differently (or if it even matters). Here is what I'm seeing:

Screen Shot 2022-01-11 at 8 28 45 AM

martinfleis: That is a reported bug affecting Safari only. Firefox or chrome are showing it fine. I'd say we let Github fix it.

jGaboardi (approved):

martinfleis:
Let's see how it will work!

PR #342, opened 2022-01-17
pytest-xdist for multicore CI

jGaboardi:
xref: pysal/spopt#211, pysal/spaghetti#661

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -61,7 +61,7 @@ jobs:
 
       - name: Test momepy
         run: |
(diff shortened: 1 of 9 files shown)


jGaboardi:
[Failures](https://github.com/pysal/momepy/runs/4835325249?check_suite_focus=true#step:6:244) all have acceptable tolerances. However, this seems to be be the first time they cropped up. Wondering is lack of precision to do with multicore schema.

martinfleis:
Can you wrap the affected tests in `pytest.approx`? We have done it elsewhere as well.

jGaboardi:
@martinfleis All precision failures are now addressed.

martinfleis:
Thanks!

PR #343, opened 2022-01-18
DOC: add clustering example from SDSC

martinfleis:
Introduce Examples section of docs and add the notebook I used during the Spatial Data Science Conference workshop a few months ago.


martinfleis:
See https://momepy--343.org.readthedocs.build/en/343/examples/clustering.html for a rendered version.

jGaboardi (approved):

PR #347, opened 2022-03-15
BUG: Fix angle computation in graph creation with dual approach

### LINKED ISSUE #346: BUG?: angular betweenness is returning strange results
### Describe the problem

Angular betweenness is apparently not correct. I think. Instead of giving us smooth result, it often returns step-like fragments. I don't know what is causing it but it is present in most of cases so it is something we are doing wrong. Somewhere. It will require a bit of debugging and finding an alternative implementation to compare to (that may be tricky).

Screenshot 2022-03-15 at 10 06 04
 

### Steps to reproduce

This is visible even in our user guide, so you can reproduce it using that. http://docs.momepy.org/en/stable/user_guide/graph/centrality.html#Angular-betweenness

### Versions of your packages

latest

### Your operating system

_No response_

### Additional context

_No response_

martinfleis:
Closes #346 

tldr: I messed up basic math

Okay, this was actually easier that I thought. When creating a graph using the dual approach, the angle between adjacent streets was computed incorrectly. As a result, when two streets formed a continuous straight line, the angle was pi, the largest possible option instead of 0. I checked how @songololo does that in `cityseer` and shamelessly stole his formula as it works (manually verified).

I have also checked the results on the same case where I found those "steps" showed in #346 and it is nice and smooth as you would expect now.

That said, I'd be happy to see how another implementation does the same thing, to avoid other surprises. 

@songololo can `cityseer` do this centrality? What we do here (see the code below) is creation of a graph where each street segment is a node and intersection between two nodes is encoded as an edge. We then measure the angle between both adjacent edges (simplified into first and last coordinates, which is _I think_ what Space Syntax refers to as Tasos implementation) and use this angle as a weight when measuring centrality. I see that you deal with angular centrality a bit differently and am not sure how exactly to be fair.

Snippet used:

```py
import geopandas
import momepy

edges = geopandas.read_file("/Users/martin/Downloads/network_prg.gpkg")
dual = momepy.gdf_to_nx(edges, approach='dual')
dual = momepy.betweenness_centrality(dual, name='angular_betweenness', mode='nodes', weight='angle')
edges_d = momepy.nx_to_gdf(dual)
```

The GPKG with the resulting values is attached. Just FYI, on my M1 MacBook the betweenness takes 3 mins 35 s on this dataset, which is sad and illustrates why I'd be happy to replace these NetworkX implementation with something faster, like `cityseer`'s numba algorithms (if we find it to be feasible). 

[network_prg.gpkg.zip](https://github.com/pysal/momepy/files/8257500/network_prg.gpkg.zip)


martinfleis:
Failures on dev are caused by https://github.com/geopandas/geopandas/issues/2282

momepy/utils.py
```diff
@@ -38,18 +38,13 @@ def unique_id(objects):
...
-    ba = [aa - bb for aa, bb in zip(a, b)]
-    bc = [cc - bb for cc, bb in zip(c, b)]
-    nba = math.sqrt(sum((x ** 2.0 for x in ba)))
-    ba = [x / nba for x in ba]
-    nbc = math.sqrt(sum((x ** 2.0 for x in bc)))
-    bc = [x / nbc for x in bc]
-    scal = sum((aa * bb for aa, bb in zip(ba, bc)))
-    angle = math.acos(round(scal, 10))
-    return angle
+    a1 = math.degrees(math.atan2(b[1] - a[1], b[0] - a[0]))
```
jGaboardi, line 45: Is there a specific reason for using `math` instead of `numpy` for these calculations?

martinfleis: Yes, it is way faster for scalars. numpy would make sense if we used arrays here, which we don't (we may refactor it, though)

```py
%%timeit
np.rad2deg(np.arctan2(1, 0))

1.26 µs ± 47.4 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

%%timeit
math.degrees(math.atan2(1, 0))

100 ns ± 3.83 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

jGaboardi (approved):

PR #350, opened 2022-03-26
TST: skip to avoid geopandas regression

martinfleis:
We have two tests failiung due to https://github.com/geopandas/geopandas/issues/2282 . Skipping to avoid CI failures. It will be fixed in geopandas ahead of 0.11.


martinfleis:
The issue with the windows env is not related. I see the same Fiona installation issue in splot.

The issue with RTD is a known bug that will be fixed in upstream.

jGaboardi (approved):

PR #351, opened 2022-04-08
BUG: fix issue with blocks within another blocks

martinfleis:
Fixing the case when one block is fully within the exterior of the other. It currently creates overlapping blocks, this should fix it.

### DIFF
--- a/momepy/elements.py
+++ b/momepy/elements.py
@@ -662,9 +662,7 @@ class Blocks:
                 buildings_c, blocks, how="left", predicate="within"
             )
         else:
-            centroids_w_bl_ID2 = gpd.sjoin(
-                buildings_c, blocks, how="left", op="within"
-            )
+            centroids_w_bl_ID2 = gpd.sjoin(buildings_c, blocks, how="left", op="within")
 
         self.buildings_id = centroids_w_bl_ID2[id_name]
 
--- a/momepy/elements.py
+++ b/momepy/elements.py
@@ -594,19 +594,6 @@ class Blocks:
     Examples
     --------
     >>> blocks = mm.Blocks(tessellation_df, streets_df, buildings_df, 'bID', 'uID')
-    Buffering streets...
-    Generating spatial index...
-    Difference...
-    Defining adjacency...
-    Defining street-based blocks...
-    Defining block ID...
-    Generating centroids...
-    Spatial join...
-    Attribute join (tesselation)...
-    Generating blocks...
-    Multipart to singlepart...
-    Attribute join (buildings)...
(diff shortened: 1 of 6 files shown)


momepy/elements.py
```diff
@@ -648,14 +635,14 @@ def __init__(self, tessellation, edges, buildings, id_name, unique_id, **kwargs)
                 buildings_c,
                 cut[[cut.geometry.name, "component"]],
                 how="left",
-                predicate="intersects",
+                predicate="within",
```
martinfleis, line 638: within is faster and with point in polygon results in the same thing

momepy/elements.py
```diff
@@ -668,30 +655,23 @@ def __init__(self, tessellation, edges, buildings, id_name, unique_id, **kwargs)
                 cells_copy.dissolve(by="component").explode().reset_index(drop=True)
             )
         blocks[id_name] = range(len(blocks))
-        blocks[blocks.geometry.name] = gpd.GeoSeries(
-            pygeos.polygons(blocks.exterior.values.data), crs=blocks.crs
-        )
```
martinfleis, line 673: I originally did this to cleanup interior after complex dissolve but it seems that GEOS got much better on it so it may not be needed. This is what was causing the issue in the first place, so if we want some cleaning procedure, it needs to be different.

momepy/elements.py
```diff
@@ -668,30 +655,23 @@ def __init__(self, tessellation, edges, buildings, id_name, unique_id, **kwargs)
...
-                buildings_c, blocks, how="left", op="intersects"
+                buildings_c, blocks, how="left", op="within"
             )
-        bl_ID_to_uID = centroids_w_bl_ID2[[unique_id, id_name]]
 
-        buildings_m = buildings[[unique_id]].merge(
-            bl_ID_to_uID, on=unique_id, how="left"
-        )
-        self.buildings_id = buildings_m[id_name]
-        self.buildings_id.index = self.buildings.index
```
martinfleis, line 690: This is just a simplification.

jGaboardi (approved):

jGaboardi:
@martinfleis Should we add a `codecov.yml` so that we don't get these failing checks from `codecov`? If yes, I can put in the PR.

martinfleis:
> Should we add a codecov.yml so that we don't get these failing checks from codecov?

Yes. We should also drop python 3.7, can you include that bump in the same one?

jGaboardi:
> > Should we add a codecov.yml so that we don't get these failing checks from codecov?
> 
> Yes. We should also drop python 3.7, can you include that bump in the same one?

Sure.

PR #352, opened 2022-04-08
CI updates

jGaboardi:
This PR:
* add a `codecov.yml` to customize coverage reporting
* drops 3.7 testing
* moves minimal testing to 3.8
* moves macOS and Windows testing to 3.10


ci/envs/37-minimal.yaml
```diff
@@ -1,18 +0,0 @@
-name: test
-channels:
-  - conda-forge
-dependencies:
-  - python=3.7
-  - geopandas=0.8
-  - libpysal=4.2.2
-  - networkx=2.3
-  - tqdm=4.27.0
```
martinfleis, line 9: The point of `minimal` is to pin the oldest supported versions of mandatory dependencies. It is likely that we should bump those as well :D, following NEP29 but I'd like to have one env with these to ensure we actually support what we claim we do. Removing it is the reason why the coverage dropped.

martinfleis: >  It is likely that we should bump those as well :D, following NEP29

Actually not, this seems to be okay. If you can just move those pins to 38-minimal, we can merge.

martinfleis (approved):
Thanks!

PR #357, opened 2022-05-19
ENH: support bi-directional roads in gdf_to_nx

kopytjuk:
Some road sections can be passed in a bidirectional manner. Either the user has to preprocess his GeoDataFrame (to add an additional LineString in a opposite direction) or we can introduce a function where based on a boolean column, a user can specify which linestrings (roads) are `oneway` or not:

image

This functionality is helpful for path finding applications in directed graphs, where the same road can be traversed in the opposite direction.

### DIFF
--- a/momepy/utils.py
+++ b/momepy/utils.py
@@ -65,7 +65,6 @@ def _generate_primal(G, gdf_network, fields, multigraph, oneway_column=None):
             key += 1
 
             if oneway_column:
-                print(row)
                 oneway = bool(getattr(row, oneway_column))
                 if not oneway:
                     G.add_edge(last, first, key=key, **attributes)
@@ -241,7 +240,9 @@ def gdf_to_nx(
 
     if approach == "primal":
         if oneway_column and ((not directed) and (not multigraph)):
-            raise ValueError("Bidirectional lines are only supported for directed multigraphs.")
+            raise ValueError(
+                "Bidirectional lines are only supported for directed multigraphs."
+            )
 
         _generate_primal(net, gdf_network, fields, multigraph, oneway_column)
 
--- a/momepy/utils.py
+++ b/momepy/utils.py
@@ -47,7 +47,7 @@ def _angle(a, b, c):
     return abs((a2 - a1 + 180) % 360 - 180)
 
 
-def _generate_primal(G, gdf_network, fields, multigraph):
+def _generate_primal(G, gdf_network, fields, multigraph, oneway_column=None):
     """
     Generate primal graph.
     Helper for gdf_to_nx.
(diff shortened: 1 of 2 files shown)


momepy/utils.py
```diff
@@ -63,6 +63,13 @@ def _generate_primal(G, gdf_network, fields, multigraph):
         if multigraph:
             G.add_edge(first, last, key=key, **attributes)
             key += 1
+
+            if oneway_column:
+                print(row)
```
jGaboardi, line 68: Is there a reason this `print` is left in?

momepy/utils.py
```diff
@@ -229,7 +239,12 @@ def gdf_to_nx(
     fields = list(gdf_network.columns)
 
     if approach == "primal":
-        _generate_primal(net, gdf_network, fields, multigraph)
+        if oneway_column and ((not directed) and (not multigraph)):
```
martinfleis, line 242: This could in theory work also for `networkx.DiGraph`, not only `MultiDiGraph`, right? Can we allow it?

momepy/utils.py
```diff
@@ -229,7 +239,12 @@ def gdf_to_nx(
...
+            raise ValueError(
```
martinfleis, line 243: Can you add a test checking the error?

martinfleis (commented):
Thanks a lot! This is surely useful. I've left two notes in the code.

If you want, you can also add this to the user guide page http://docs.momepy.org/en/stable/user_guide/graph/convert.html. (not strictly needed though).

One more thing we should certainly note - when you do this, `gdf_to_nx` -> `nx_to_gdf` will not roundtrip as the gdf coming from the bidirectional graph will contain duplicated geometries. That needs to be documented and we may even think about a way of dealing with this when needed.

momepy/utils.py
kopytjuk: You are right, removed the restriction.

kopytjuk:
> Thanks a lot! This is surely useful. I've left two notes in the code.
> 
> If you want, you can also add this to the user guide page http://docs.momepy.org/en/stable/user_guide/graph/convert.html. (not strictly needed though).
> 
> One more thing we should certainly note - when you do this, `gdf_to_nx` -> `nx_to_gdf` will not roundtrip as the gdf coming from the bidirectional graph will contain duplicated geometries. That needs to be documented and we may even think about a way of dealing with this when needed.

Agreed, I added a note in the docstring of the function that with oneway-option `nx_to_gdf(gdf_to_nx(gdf, directed=True, oneway_column="oneway"))` an extended geodataframe is returned.

Regarding the user guide, I would prefer a separate PR for that, I guess writing it and generating images will be more work than this PR - if it is OK for you, I would leave this PR as it is.

kopytjuk:
Hello again @martinfleis, am I missing something?

martinfleis:
Sorry, long backlog of open source tasks :D. I think it is ready to go. I'll give it another look later today and merge unless I find some issue.

jGaboardi (approved):

martinfleis:
Thanks a lot @kopytjuk!

PR #371, opened 2022-07-31
ENH: geometry-based simplification of roundabouts

gregmaya:
I'm sure there might be corrections to be made (including one BUG FIX that I found whilst doing the test!) but I believe submitting is helpful to move forward.


martinfleis:
Hey, I have renamed the PR and resolved the merge conflict. 

1) When creating a PR, try to always ensure that it is based on the latest version of `main`. When it is a simple one, make sure that your local repo is up to date and then make a new branch from main. When it is a larger one that takes time, do the same in the beginning and then merge main to the branch before creating a PR

```
git fetch upstream
git merge upstream/main
```

2) make sure you follow the [style guide](http://docs.momepy.org/en/stable/contributing.html#formatting-the-code). As you can see from the linting check, it does not follow `black` at the moment. Optionally (but recommended), you can setup [pre-commit hooks](https://pre-commit.com/) to automatically run black when you make a git commit. 

```
python -m pip install pre-commit
```
From the root of the momepy repository, you should then install the pre-commit included in GeoPandas:
```
pre-commit install
```
Then black will be run automatically each time you commit changes. You can skip these checks with `git commit --no-verify`.
 

I'll check the actual contents later ;).

gregmaya:
Wow.! Ok definitely PRs are a science in itself! ..
Thanks for the explanations... although some still sound foreign to me I'll have your recommendations to start Googling. 
Many thanks @martinfleis !

gregmaya:
sorry I marked the PR as 'ready for review' by mistake! (not sure how to undo 😅)

momepy/preprocessing.py
```diff
@@ -740,3 +748,240 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+                if hdist > rab_diameter :
+                    adj_long_polys.append(rrow[0]) #if hausdorff_distance is larger than the diameter add to list
+           
+        rab_plus = rab_adj.drop(adj_long_polys)
+        
+    else:
+        rab['index_right'] = rab.index
+        rab_plus = rab
+    
+    rab_plus = rab_plus[['geometry', 'index_right']] #only keeping relevant fields
```
martinfleis, line 797: You can't be certain you will always get "geometry" and "index_right" columns unless you sanitise the given gdf yourself. What a user may pass can have custom geometry columns name and a named index and this will no longer work.

momepy/preprocessing.py
```diff
@@ -740,3 +748,240 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+
+def _rabs_center_points(gdf, center_type = 'centroid'):
+    """
+    From a selection of round abouts, returns an aggregated GeoDataFrame 
+    per round about with extra column with center_type. 
+    """
+    #creating a multipolygon per RAB (as opposed to dissolving) of the entire composition of the RAB
+    # temporary DataFrame where geometries is the array of pygeos geometries
+    # for aggregation to work properly
+    tmp = pd.DataFrame(gdf.copy())
```
martinfleis, line 809: Can you make a note here that this is a temporary hack until shapely 2.0 is out?

momepy/preprocessing.py
```diff
@@ -740,3 +748,240 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+    tmp['geometry'] = tmp.geometry.values.data
+
+    rab_plus = gpd.GeoDataFrame(
+            tmp.groupby('index_right').geometry.apply(pygeos.multipolygons).rename('geometry'), crs= gdf.crs)
+    # verifying that all geometries are valid
+    rab_plus['geometry'] = rab_plus.apply(lambda row : make_valid(row.geometry) if not row.geometry.is_valid else row.geometry, axis = 1)
```
martinfleis, line 815: if you are using pygeos here anyway, use their vectorised versions of `make_valid` and `is_valid`. This will be extremely slow due to 1) a loop, 2) a conversion between underlying pygeos geometry and a shapely geometry within the loop.

momepy/preprocessing.py
```diff
@@ -740,3 +748,240 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+    
+    if center_type == 'centroid' :
+        #geometry centroid of the actual circle
+        rab_plus['center_pt'] = gdf[gdf.index == gdf.index_right].geometry.centroid
+    
+    elif center_type == 'mean':
+        # mean geometry
+        ls_xy = [g.exterior.coords.xy for g in rab_plus.geometry] #extracting the points
```
martinfleis, line 823: Again, you can get an array of coordinates with pygeos without a loop.

momepy/preprocessing.py
```diff
@@ -740,3 +748,240 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+        
+        ls_xy = [point for polygon in multipolygon for point in polygon.exterior.coords[:-1]]
```
martinfleis, line 825: I have no clue what this does :D.

momepy/preprocessing.py
```diff
@@ -740,3 +748,240 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+        
+        mean_pts = [geometry.Point(np.mean(xy[0]),np.mean(xy[1])) for xy in ls_xy]
```
martinfleis, line 827: get arrays and create points with `points_from_xy` or pygeos. always try to use a vectorised solution if there is one.

momepy/preprocessing.py
```diff
@@ -740,3 +748,240 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+        rab_plus['center_pt'] = gpd.GeoSeries( data = mean_pts, 
+                                              index = rab_plus.index, 
+                                              crs = edges.crs)
+    
+    # centerpoint of minimum_bounding_circle
+    # minimun_bounding_circle() not available in Shapely 1.8.2 but only in 'latest'
+    # --> https://shapely.readthedocs.io/en/latest/constructive.html
+    # current environment has 1.8.2
```
martinfleis, line 835: Can you add that it will be available in 2.0?

momepy/preprocessing.py
```diff
@@ -740,3 +748,240 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+        result_idx = gs1.sjoin(gs2 , predicate = 'covered_by').index
+        coins_filter_result.extend(result_idx)
+    
+    incoming_many_reduced = incoming_many.loc[coins_filter_result]
+    return incoming_many_reduced
+
+def _selecting_incoming_lines (rab_plus, edges, angle_threshold=0):
+    """Selecting only the lines that are touching but not covered by
+    the ``rab_plus``.
+    If more than one LineString is incoming to ``rab_plus``, COINS algortthm
```
martinfleis, line 872:
```suggestion
    If more than one LineString is incoming to ``rab_plus``, COINS algorithm
```

momepy/preprocessing.py
```diff
@@ -740,3 +748,240 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+    is used to select the line to be extended further.
+    """
+    # Feels a bit combersome ... Ideally there would be a DISJOINT predicate
```
martinfleis, line 875: I am not sure if disjoint is what you are looking for here.

> Returns True if A and B do not share any point in space. Disjoint implies that overlaps, touches, within, and intersects are False.

momepy/preprocessing.py
```diff
@@ -740,3 +748,240 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+    incoming = edges.sjoin(rab_plus , predicate = 'touches')
+    incoming.rename(columns ={'index_right':'index_rab_plus'}, inplace = True )
+    idx_drop =  incoming.sjoin(rab_plus, predicate = 'covered_by').index
+    incoming.drop(idx_drop, axis=0, inplace =True)
```
martinfleis, line 879: You could create your own query on `.sindex` here and make the filter before doing the actual join.

momepy/preprocessing.py
```diff
@@ -740,3 +748,240 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+    for i, row in incoming.iterrows() :
+        if row.dist_fisrt_pt < row.dist_last_pt :
+            lines.append(geometry.LineString([row.first_pt, row.center_pt]))
+        else :
+            lines.append(geometry.LineString([row.last_pt, row.center_pt]))
+    incoming['line'] = gpd.GeoSeries(lines, index=incoming.index ,crs= edges.crs)
+
+    #checking in there are more than one incoming lines arriving to the same point
+    #which would create several new lines
+    incoming['line_wkt'] = incoming.line.apply(lambda x : x.wkt)
```
martinfleis, line 896:
```suggestion
    incoming['line_wkt'] = incoming.line.to_wkt()
```

momepy/preprocessing.py
```diff
@@ -740,3 +748,240 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+
+def _coins_filtering_many_incoming(incoming_many, angle_threshold=0):
+    """
+    Used only for the cases when more than one incoming line touches the
+    roundabout.
+    """
+    # figuring out which one needs to be extended and retain attributes
+    coins_filter_result = []
+    # For each new connection, evaluate COINS and selecet the group from which the new line belongs
+    for g, x in incoming_many.groupby('line_wkt'):
```
martinfleis, line 847: you are already doing `groupby` over `line_wkt` before you call this function. Can you pass the groupby object directly instead of doing another groupby?

momepy/preprocessing.py
```diff
@@ -740,3 +748,240 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+        gs = gpd.GeoSeries( pd.concat([x.geometry, x.line]), crs= incoming_many.crs )
+        gdf = gpd.GeoDataFrame(geometry = gs)
+        gdf.drop_duplicates(inplace=True)
+
+        coins = COINS(gdf, angle_threshold=angle_threshold)
+        stroke_gdf = coins.stroke_gdf()
+
+        orig_geom_join = stroke_gdf.sjoin(gpd.GeoDataFrame(geometry = x.line), predicate= 'covers' )
```
martinfleis, line 855: Can we instead use `coins.stroke_attribute`? We could avoid the sjoin.

momepy/preprocessing.py
```diff
@@ -740,3 +748,240 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+def _ext_lines_to_center(edges, incoming_all, rab_plus):
+    """Extends the Linestrings geometrie to the centerpoint defined by _rabs_center_points.
+    Also deleted the lines that originally defined the roundabout.
+    
+    Returns
+    -------
+    GeoDataFrame
+        GeoDataFrame of with updated geometry
+    """
+    ## this is causing a warning too for Shapely 2.0 --> Convert the '.coords' to a numpy array
```
martinfleis, line 919: That is fine, you can ignore it.

momepy/preprocessing.py
```diff
@@ -740,3 +748,240 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+    incoming_all['geometry'] = incoming_all.apply(lambda row: linemerge([row.geometry, row.line]), axis =1)
+
+    # deleting the original round about edges
+    idx_out = edges.sjoin(rab_plus, predicate= 'covered_by', how='inner').index
```
martinfleis, line 923: You have them selected above, you know their ID's. Try to reuse that and drop them based on the index instead of doing another spatial join. Those are expensive.

momepy/datasets/__init__.py
```diff
@@ -3,7 +3,7 @@
 __all__ = ["available", "get_path"]
 
 _module_path = os.path.dirname(__file__)
-available = ["bubenec", "tests"]
+available = ["bubenec", "tests", "mad_test_rabs"]
```
martinfleis, line 6: Can you add it as another layer of `tests.gpkg` instead of adding a new one?

martinfleis (commented):
Hey, a first pass of comments in the code.

momepy/preprocessing.py
```diff
@@ -740,3 +748,240 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+        #selecting the adjacent areas that are of smaller than itself
+        rab_adj = gpd.sjoin(gdf, rab, predicate = 'intersects')
+        rab_adj = rab_adj[rab_adj.area_right >= rab_adj.area_left]
+        
+        adj_long_polys = []
+        #iterate through original rabs
+        for row in rab.itertuples() :
+            #creating a distance (diameter) for limiting the max dist of adjacent
+            minx, _ , maxx, _ = row.geometry.bounds 
+            rab_diameter = maxx - minx
```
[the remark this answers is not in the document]
gregmaya: I used this as a proxy of the radius of the roundabout. It could also be `maxy - miny`, but assuming most roundabouts are close to a perfect circle those the differences on X and Y axis should be very similar. 
Ideally, one would take the radius from a minimum bounding circle, but I know this is a bit more challenging. 
I will incorporate a small upgrade to this, though.
minx, _ , maxx, _ = row.geometry.bounds 
```suggestion
            minx, maxy , maxx, miny = row.geometry.bounds 
            if (maxx - minx) >= (maxy - miny):
                rab_diameter = maxx - minx
            else:
                rab_diameter = maxy - miny
```

momepy/preprocessing.py
gregmaya: humm good point 👍 
I'm not sure how to deal with the 'index_right' just yet (possibly by reseting the index at the very begining?!)..
but for now I'll implement `geom_col = rab_plus.geometry.name` for the geometry column

martinfleis: You can reset index but you should reconstruct it afterwards, to return gdf in a same structure as it comes in.

momepy/preprocessing.py
```diff
@@ -740,3 +748,240 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+            
+            #creating the subgroup that could be affected
+            rab_group = rab_adj[rab_adj.index_right==row[0]]
+            for rrow in rab_group.itertuples():
+                hdist = row.geometry.hausdorff_distance(rrow.geometry)
+
+                if hdist > rab_diameter :
+                    adj_long_polys.append(rrow[0]) #if hausdorff_distance is larger than the diameter add to list
```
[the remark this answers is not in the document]
gregmaya: I need to come back to this to see if I can implement @gregmaya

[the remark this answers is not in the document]
martinfleis: Feel free to skip it now and add a # TODO comment there.

momepy/preprocessing.py
gregmaya: Not sure I quite follow what you mean by this. 
I will keep the comment open to review later

momepy/preprocessing.py
gregmaya: I tried this, but was somehow, unsuccessful because the earlier grouping was over a different gdf. 
Need to come back to this. #TODO

momepy/preprocessing.py
```diff
@@ -15,14 +15,20 @@
...
+from .coins import COINS
+
 
 __all__ = [
     "preprocess",
     "remove_false_nodes",
     "CheckTessellationInput",
     "close_gaps",
     "extend_lines",
+    "roundabout_simpl",
```
martinfleis, line 31:
```suggestion
    "roundabout_simplification",
```

docs/api.rst
```diff
@@ -150,6 +150,7 @@ preprocessing
    extend_lines
    remove_false_nodes
    preprocess
+   roundabout_simpl
```
martinfleis, line 153:
```suggestion
   roundabout_simplification
```

tests/test_preprocess.py
```diff
@@ -124,3 +125,37 @@ def test_extend_lines(self):
         ext5 = mm.extend_lines(gdf, 2)
         assert ext5.length.sum() > gdf.length.sum()
         assert ext5.length.sum() == pytest.approx(6.2, rel=1e-3)
+
+    def test_roundabout_simplification(self):
+        test_file_path2 = mm.datasets.get_path("tests")
+        gdf_streets = gpd.read_file(
+            test_file_path_rabs, layer="test_rabs"
```
martinfleis, line 132:
```suggestion
            test_file_path2, layer="test_rabs"
```

momepy/preprocessing.py
```diff
@@ -740,3 +746,296 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+    per round about with extra column with center_type.
+    """
+    # creating a multipolygon per RAB (as opposed to dissolving) of the entire composition of the RAB
+    # temporary DataFrame where geometry is the array of pygeos geometries
+    tmp = pd.DataFrame(gdf.copy())  # temporary hack until shapely 2.0 is out
+    tmp["geometry"] = tmp.geometry.values.data
+
+    pygeos_geoms = (
+        tmp.groupby("index_right")
+        .geometry.apply(pygeos.multipolygons)
```
jGaboardi, line 826: @martinfleis Would it be more efficient to be using `map` here vs. `apply`?

martinfleis: Possibly

momepy/preprocessing.py
```diff
@@ -740,3 +746,331 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+        # .stroke_gdf()
+        stroke_gdf = coins.stroke_gdf()
+        if GPD_10:
+            orig_geom_join = stroke_gdf.sjoin(
+                gpd.GeoDataFrame(geometry=x.line), predicate="covers"
+            )
+        else:
+            orig_geom_join = stroke_gdf.sjoin(
+                gpd.GeoDataFrame(geometry=x.line), op="covers"
+            )
```
martinfleis, line 888:
```suggestion
            orig_geom_join = gpd.sjoin(stroke_gdf,
                gpd.GeoDataFrame(geometry=x.line), op="covers"
            )
```
You need to use a function for older geopandas, the method is newer. Also note that this suggestion does not follow `black`.

momepy/preprocessing.py
```diff
@@ -740,3 +746,331 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+
+        gs1 = gpd.GeoSeries(x.geometry, crs=incoming_many.crs)
+        gs1 = gpd.GeoDataFrame(geometry=gs1)
+
+        # select the the line that's covered by the joined line returned by COINS
+        # one could consider using pygeos shared_paths(a, b) # TODO
+        if GPD_10:
+            result_idx = gs1.sjoin(gs2, predicate="covered_by").index
+        else:
+            result_idx = gs1.sjoin(gs2, op="covered_by").index
```
martinfleis, line 902:
```suggestion
            result_idx = gpd.sjoin(gs1, gs2, op="covered_by").index
```

Same as before.

You could replace the sjoin here with an `sindex` query though, to avoid the overhead of sjoin but I'd need to run this to actually write that snippet, so feel free to ignore this comment.

momepy/preprocessing.py
```diff
@@ -740,3 +746,331 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+    """Selecting only the lines that are touching but not covered by
+    the ``rab_plus``.
+    If more than one LineString is incoming to ``rab_plus``, COINS algorithm
+    is used to select the line to be extended further.
+    """
+    # selecting the lines that are touching but not covered by
+    if GPD_10:
+        touching = edges.sjoin(rab_multipolygons, predicate="touches")
+    else:
+        touching = edges.sjoin(rab_multipolygons, op="touches")
```
martinfleis, line 920:
```suggestion
        touching = gpd.sjoin(edges, rab_multipolygons, op="touches")
```

momepy/preprocessing.py
```diff
@@ -740,3 +746,331 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+
+    if GPD_10:
+        idx_drop = edges.sjoin(rab_multipolygons, predicate="covered_by").index
+    else:
+        idx_drop = edges.sjoin(rab_multipolygons, op="covered_by").index
```
martinfleis, line 925:
```suggestion
        idx_drop = gpd.sjoin(edges, rab_multipolygons, op="covered_by").index
```

momepy/preprocessing.py
gregmaya: not sure I fully understand what you mean... but would be curious to see what you suggest to still maintain the list of indexes as an output of the  function (which is later used to remove the original lines from the edges gdf)

gregmaya:
Hoping the auto-BLACK pre-commit action is actually working. 🧐

momepy/preprocessing.py
martinfleis: Take this example:

```py
In [1]: import geopandas
world = geopa                                                                  
In [2]: world = geopandas.read_file(geopandas.datasets.get_path('naturalearth_lowres'))

In [3]: cities = geopandas.read_file(geopandas.datasets.get_path('naturalearth_cities'))
```

This is your way - do the whole sjoin and then extract only the index to know what should be dropped.

```py
In [4]: sjoin_idx = cities.sjoin(world).index

In [5]: sjoin_idx
Out[5]: 
Int64Index([  0,   1, 192,   2, 184,   3,   8,  11, 199,  12,
            ...
            188, 190, 191, 193, 194, 195, 196, 197, 198, 200],
           dtype='int64', length=172)

```
Under the hood, geopandas calls `.sindex.query_bulk` to know which geometries intersect each other (or are covered by if `predicate="covered_by"`) and does a lot of other things to construct the final GeoDataFrame. But we're not interested in those other things.

So you can call just the sindex yourself:

```py
In [6]: world_idx, cities_idx = cities.sindex.query_bulk(world.geometry, predicate="intersects")
```
We get two arrays out, representing pairs of geometries in each gdf that fulfil the predicate. Note, that these are integer (positional) indices. 

See that the one for `cities` has the same length as yours and since `cities.index` is a default RangeIndex, hence its values are the same as a position, all values actually match, they are just sorted in a different order. 
```

In [7]: len(cities_idx)
Out[7]: 172

In [8]: all([x in sjoin_idx for x in cities_idx])
Out[8]: True
```

If the original index wasn't the default one, you get the result you want like this:

```py
In [9]: cities.index.take(cities_idx)
Out[9]: 
Int64Index([ 91, 144, 101, 186,  96, 165,  59, 195, 182, 200,
            ...
            123,  48,  13, 120,  25, 102,  19,  21,  12,  15],
           dtype='int64', length=172)
```

The reason to do this is to avoid the overhead of sjoin. This is not the best test case, but you can check the difference in performance yourself.

```py
In [10]: %timeit cities.sjoin(world).index
4.08 ms ± 11.3 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [11]: %timeit cities.sindex.query_bulk(world.geometry, predicate="intersects")
425 µs ± 482 ns per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

martinfleis:
> Hoping the auto-BLACK pre-commit action is actually working. 

It looks it is :).

momepy/preprocessing.py
gregmaya: I just included this but `query_bulk` doesn't have the predicate `'covered_by'` so I ended up changing some things to still get the desired output...

```
`edges_idx,` rabs_idx = rab_multipolygons.buffer(1).sindex.query_bulk(
            edges.geometry, predicate="within")
```

 I couldn't find a way to use any of the predicates without using the buffer

martinfleis: It does have all the predicates. Without that we couldn't expose it in sjoin. See https://pygeos.readthedocs.io/en/stable/strtree.html#pygeos.strtree.STRtree.query_bulk

gregmaya: of course, that makes sense since it's under the hood of spatial joins... 
I was looking here instead: https://geopandas.org/en/stable/docs/reference/api/geopandas.sindex.SpatialIndex.query_bulk.html

This is what I keep testing and still get different results than doing the original `sjoin`

> rabs_idx, edges_idx = edges.sindex.query_bulk(rab_multipolygons.geometry, predicate= "covered_by")
> idx_drop = edges.index.take(edges_idx)
-> `idx_drop`  is empty in this case

> idx_drop = edges.sjoin(rab_multipolygons, predicate="covered_by").index
-> `idx_drop` actually has the roundabout lines (excluding the incoming lines)

Any idea what could be happening? (I feel it has to be an oversight from my side)

martinfleis: It is not the most intuitive but the sindex query is working in an opposite direction to sjoin

- sindex: return _edges_ that are _"covered_by"_ by _rab_multipolygons.geometry_
- sjoin: join _rab_multipolygons_ to _edges_ if _rab_multipolygons_ are _"covered by"_ by _edges_

So the query needs to be opposite.

jGaboardi:
`"touches"` is not a valid predicate argument in [geopandas v0.8.2](https://geopandas.org/en/v0.8.2/reference/geopandas.sjoin.html?highlight=spatial%20join), which is the [version being pulled in](https://github.com/pysal/momepy/runs/7793755697?check_suite_focus=true#step:4:11) for the 3.8-minimal CI run, resulting in the current [failure](https://github.com/pysal/momepy/runs/7793755697?check_suite_focus=true#step:6:194).

martinfleis:
I wouldn't bother trying to make it work for 0.8.2 to be fair. There's too many constraints.

momepy/preprocessing.py
gregmaya: The latest update now includes `coins.stroke_attribute` following the correction done last week by @amorfinv ! 
Much cleaner, indeed!

martinfleis:
@gregmaya give us a shout if you want another round of reviews.

gregmaya:
> @gregmaya give us a shout if you want another round of reviews.

A review would be appreciated, yes! @martinfleis 
I really want to move to a selection algorithm (which I've already started some explorations)... but obviously leaving something working here is crucial. 
For the sake of future improvements on this `roundabout_simplification` I'm going to list a few ideas to include/solve:

1. The current selection part of the algorithm is classifying many false positives (i.e. all those roundabouts that are 'cut in half' by streets continuing their original direction. 
2. Include an extra column on the output with a label that highlights the streets that were extended and were 'incoming to a roundabout'.
3. Create an optional return with the deleted streets as a separate gdf. 
(points 2 and 3 could be implemented in all the functions that aim at network simplification)
4. Overall improvements of inefficient coding.

momepy/preprocessing.py
```diff
@@ -740,3 +746,313 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+    above the Circular Compaactness threshold.
+
+    Return
+    ________
+    GeoDataFrames : round abouts and adjacent polygons
+    """
+    # calculate parameters
+    gdf = gdf.copy()
+    gdf["area"] = gdf.geometry.area
+    gdf["circom"] = CircularCompactness(gdf, "area").series
```
martinfleis, line 772: Strictly speaking, you don't need to assign these as columns because those boolean filters you are creating below will work as well with the series' directly. Then you can probably avoid the copy above.

momepy/preprocessing.py
```diff
@@ -740,3 +746,313 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+
+def _rabs_center_points(gdf, center_type="centroid"):
+    """
+    From a selection of roundabouts, returns an aggregated GeoDataFrame
+    per round about with extra column with center_type.
+    """
+    # creating a multipolygon per RAB (as opposed to dissolving) of the entire
+    # composition of the RAB
+    # temporary DataFrame where geometry is the array of pygeos geometries
+    tmp = pd.DataFrame(gdf.copy())  # temporary hack until shapely 2.0 is out
```
martinfleis, line 825: Why?

martinfleis: Ahh, I see why. Can you expand this comment to explain the hack?

momepy/preprocessing.py
```diff
@@ -740,3 +746,313 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+            tmps = coords[idxs == i]
+            target_idx = rab_multipolygons.index[i]
+            means[target_idx] = Point(tmps.mean(axis=0))
+
+        rab_multipolygons["center_pt"] = gpd.GeoSeries(means, crs=gdf.crs)
+
+    # centerpoint of minimum_bounding_circle
+    # minimun_bounding_circle() should be available in Shapely 2.0. Implementation still
+    # pending.
+    # current environment has 1.8.2
```
martinfleis, line 858: Add `# TODO` string here and let it be.

momepy/preprocessing.py
```diff
@@ -740,3 +746,313 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+    """
+    Extends the Linestrings geometrie to the centerpoint defined by
+    _rabs_center_points. Also deleted the lines that originally defined the roundabout.
+
+    Returns
+    -------
+    GeoDataFrame
+        GeoDataFrame of with updated geometry
+    """
+    # this can most likely be vectorized with pygeos.line_merge()!! #TODO
```
martinfleis, line 963: Not really, leave it.

martinfleis (commented):
Almost there!

docs/api.rst
martinfleis, line 153:
```suggestion
   roundabout_simplification
```

martinfleis:
If you pass a dataframe with MultiLinestrings, it fails.

```py
---------------------------------------------------------------------------
NotImplementedError                       Traceback (most recent call last)
Untitled-1.ipynb Cell 4 in <cell line: 1>()
----> 1 simplified = momepy.roundabout_simplification(df)

File ~/Git/momepy/momepy/preprocessing.py:1053, in roundabout_simplification(edges, polys, circom_threshold, area_threshold, include_adjacent, center_type, angle_threshold)
   1046 rab = _selecting_rabs_from_poly(
   1047     polys,
   1048     circom_threshold=circom_threshold,
   1049     area_threshold=area_threshold,
   1050     include_adjacent=include_adjacent,
   1051 )
   1052 rab_multipolygons = _rabs_center_points(rab, center_type=center_type)
-> 1053 incoming_all, idx_drop = _selecting_incoming_lines(
   1054     rab_multipolygons, edges, angle_threshold=angle_threshold
   1055 )
   1056 output = _ext_lines_to_center(edges, incoming_all, idx_drop)
   1058 return output

File ~/Git/momepy/momepy/preprocessing.py:915, in _selecting_incoming_lines(rab_multipolygons, edges, angle_threshold)
    912 incoming = touching.loc[ls]
    914 # figuring out which ends of incoming edges needs to be connected to the center_pt
--> 915 incoming["first_pt"] = incoming.geometry.apply(lambda x: Point(x.coords[0]))
    916 incoming["dist_fisrt_pt"] = incoming.center_pt.distance(incoming.first_pt)
    917 incoming["last_pt"] = incoming.geometry.apply(lambda x: Point(x.coords[-1]))

File ~/Git/geopandas/geopandas/geoseries.py:643, in GeoSeries.apply(self, func, convert_dtype, args, **kwargs)
    641 @doc(pd.Series)
    642 def apply(self, func, convert_dtype=True, args=(), **kwargs):
--> 643     result = super().apply(func, convert_dtype=convert_dtype, args=args, **kwargs)
    644     if isinstance(result, GeoSeries):
    645         if self.crs is not None:

File ~/mambaforge/envs/geo_dev/lib/python3.10/site-packages/pandas/core/series.py:4433, in Series.apply(self, func, convert_dtype, args, **kwargs)
   4323 def apply(
   4324     self,
   4325     func: AggFuncType,
   (...)
   4328     **kwargs,
   4329 ) -> DataFrame | Series:
   4330     """
   4331     Invoke function on values of Series.
   4332 
   (...)
   4431     dtype: float64
   4432     """
-> 4433     return SeriesApply(self, func, convert_dtype, args, kwargs).apply()

File ~/mambaforge/envs/geo_dev/lib/python3.10/site-packages/pandas/core/apply.py:1088, in SeriesApply.apply(self)
   1084 if isinstance(self.f, str):
   1085     # if we are a string, try to dispatch
   1086     return self.apply_str()
-> 1088 return self.apply_standard()

File ~/mambaforge/envs/geo_dev/lib/python3.10/site-packages/pandas/core/apply.py:1143, in SeriesApply.apply_standard(self)
   1137         values = obj.astype(object)._values
   1138         # error: Argument 2 to "map_infer" has incompatible type
   1139         # "Union[Callable[..., Any], str, List[Union[Callable[..., Any], str]],
   1140         # Dict[Hashable, Union[Union[Callable[..., Any], str],
   1141         # List[Union[Callable[..., Any], str]]]]]"; expected
   1142         # "Callable[[Any], Any]"
-> 1143         mapped = lib.map_infer(
   1144             values,
   1145             f,  # type: ignore[arg-type]
   1146             convert=self.convert_dtype,
   1147         )
   1149 if len(mapped) and isinstance(mapped[0], ABCSeries):
   1150     # GH#43986 Need to do list(mapped) in order to get treated as nested
   1151     #  See also GH#25959 regarding EA support
   1152     return obj._constructor_expanddim(list(mapped), index=obj.index)

File ~/mambaforge/envs/geo_dev/lib/python3.10/site-packages/pandas/_libs/lib.pyx:2870, in pandas._libs.lib.map_infer()

File ~/Git/momepy/momepy/preprocessing.py:915, in _selecting_incoming_lines.<locals>.<lambda>(x)
    912 incoming = touching.loc[ls]
    914 # figuring out which ends of incoming edges needs to be connected to the center_pt
--> 915 incoming["first_pt"] = incoming.geometry.apply(lambda x: Point(x.coords[0]))
    916 incoming["dist_fisrt_pt"] = incoming.center_pt.distance(incoming.first_pt)
    917 incoming["last_pt"] = incoming.geometry.apply(lambda x: Point(x.coords[-1]))

File ~/mambaforge/envs/geo_dev/lib/python3.10/site-packages/shapely/geometry/base.py:954, in BaseMultipartGeometry.coords(self)
    952 @property
    953 def coords(self):
--> 954     raise NotImplementedError(
    955         "Multi-part geometries do not provide a coordinate sequence")

NotImplementedError: Multi-part geometries do not provide a coordinate sequence
```

We should probably capture it and raise an informative error pointing towards `explode()` before simplification.

martinfleis:
There are some warnings emitted by pandas.

```py
/Users/martin/Git/momepy/momepy/preprocessing.py:802: SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
  rab_adj.hdist.loc[g.Index] = hdist
/Users/martin/Git/momepy/momepy/preprocessing.py:802: SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
  rab_adj.hdist.loc[g.Index] = hdist
/Users/martin/mambaforge/envs/geo_dev/lib/python3.10/site-packages/pandas/core/dtypes/cast.py:122: ShapelyDeprecationWarning: The array interface is deprecated and will no longer work in Shapely 2.0. Convert the '.coords' to a numpy array instead.
  arr = construct_1d_object_array_from_listlike(values)
```

Can you try to avoid them?

momepy/preprocessing.py
```diff
@@ -740,3 +746,326 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+        is above the threshold.
+
+    Returns
+    -------
+    GeoDataFrame
+        GeoDataFrame with an updated geometry and an additional column labeling modified edges.
+    """
+    assert (
+        len(edges[edges.geom_type != "LineString"]) == 0
+    ), "Only LineString geometries are allowed. Try explode() when using MultiLineString."
```
martinfleis, line 1056:
```suggestion
    if len(edges[edges.geom_type != "LineString"]) > 0:
        raise TypeError(
            "Only LineString geometries are allowed. Try using the `explode()` method to explode MultiLineStrings."
        )
```

Instead of using `assert`, raise an error. (it probably needs to be blacken)

momepy/preprocessing.py
```diff
@@ -740,3 +746,327 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+    """
+    # calculate parameters
+    gdf.loc[:, "area"] = gdf.geometry.area
+    circom_serie = CircularCompactness(gdf, "area").series
+
+    # selecting round about polygons based on compactness
+    rab = gdf[circom_serie > circom_threshold]
+    # exclude those above the area threshold
+    area_threshold_val = gdf.area.quantile(area_threshold)
+    rab = rab[rab["area"] < area_threshold_val]
```
martinfleis, line 777: Line 770 assigns a new column to gdf, which may be unwanted if the user passed it themselves. Try this (untested):

```suggestion
    area = gdf.area
    circom = CircularCompactness(gdf, area).series
    
    # selecting round about polygons based on compactness
    mask = circom_serie > circom_threshold
    rab  = gdf[mask]
    # exclude those above the area threshold
    area_threshold_val = area.quantile(area_threshold)
    rab = rab[area[mask] < area_threshold_val]
```

momepy/preprocessing.py
```diff
@@ -740,3 +746,327 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+
+def _rabs_center_points(gdf, center_type="centroid"):
+    """
+    From a selection of roundabouts, returns an aggregated GeoDataFrame
+    per round about with extra column with center_type.
+    """
+    # creating a multipolygon per RAB (as opposed to dissolving) of the entire
+    # composition of the RAB
+    # temporary DataFrame where geometry is the array of pygeos geometries
+    # Hack until shapely 2.0 is out.
```
martinfleis, line 824:
```suggestion
    # Hack until shapely 2.0 is out.
    # TODO: replace pygeos with shapely 2.0
```

it is good to keep `# TODO` in there as I sometimes look for these and fix them.

momepy/preprocessing.py
```diff
@@ -740,3 +746,327 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+    incoming_all["geometry"] = incoming_all.apply(
+        lambda row: linemerge([row.geometry, row.line]), axis=1
+    )
+
+    # deleting the original round about edges
+    new_edges = edges.drop(idx_out, axis=0)
+
+    # creating a unique label for returned gdf
+    ls = ["rab_" + x for x in incoming_all.index_right.apply(str)]
+    incoming_label = pd.Series(ls, index=incoming_all.index)
```
martinfleis, line 973:
```suggestion
    incoming_label = "rab_" + incoming_all.index_right.astype(str)
```

this is optimised in pandas, list comprehension will be slower

gregmaya: I knew there had to be something better! thanks

momepy/preprocessing.py
gregmaya: I did try that earlier. 
The reason it's useful to have it as a column is that a bit further (line 792) I use the column to define adjacent_polys.
But I will create an `if` clause for the user to provide the column name in case it already exists.

martinfleis: The line 792 is not using this column. Since you are calling it as `gdf.area` and not `gdf['area']`, the geopandas attribute measuring the area has a precedence.

gregmaya: ups ! just saw your (quick) reply.
I submitted a quick enhancement to accommodate for `area_col` as input. 

What was happening on line 792 (now 803) is that I was comparing the areas of the possible `rabs` against the areas of the possible `adjacent_polys` after the spatial join.

let me know it this too cumbersome and I can revert to something different

tests/test_preprocess.py
```diff
@@ -124,3 +125,33 @@ def test_extend_lines(self):
...
         assert ext5.length.sum() > gdf.length.sum()
         assert ext5.length.sum() == pytest.approx(6.2, rel=1e-3)
+
+    def test_roundabout_simplification(self):
+        test_file_path2 = mm.datasets.get_path("tests")
+        gdf_streets = gpd.read_file(test_file_path2, layer="test_rabs")
+        plgns = polygonize(gdf_streets.geometry)
+        gdf_polys = gpd.GeoDataFrame(geometry=[g for g in plgns], crs=gdf_streets.crs)
+
+        # test with basic params
```
jGaboardi, line 135: Is this comment needed?

tests/test_preprocess.py
```diff
@@ -124,3 +125,33 @@ def test_extend_lines(self):
         ext5 = mm.extend_lines(gdf, 2)
         assert ext5.length.sum() > gdf.length.sum()
         assert ext5.length.sum() == pytest.approx(6.2, rel=1e-3)
+
+    def test_roundabout_simplification(self):
```
jGaboardi, line 129: Would it make more sense to break this one test of 5 checks into 5 tests? Something as simple as `test_roundabout_simplification_1()`,   `test_roundabout_simplification_2()`, etc.

The data reads an polygonize can still be preformed once in the `setup_method()`.

momepy/preprocessing.py
```diff
@@ -740,3 +746,343 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+    """
+    From a GeoDataFrame of polygons, returns a GDF of polygons that are
+    above the Circular Compaactness threshold.
+
+    Return
+    ________
+    GeoDataFrames : round abouts and adjacent polygons
+    """
+    # calculate parameters
+    # calculate parameters
```
jGaboardi, line 774: Remove one of the comments. See also lines 778, 830, and 980.

momepy/preprocessing.py
```diff
@@ -740,3 +746,343 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+def _selecting_rabs_from_poly(
+    gdf,
+    area_col="area",
+    circom_threshold=0.7,
+    area_threshold=0.85,
+    include_adjacent=True,
+):
+    """
+    From a GeoDataFrame of polygons, returns a GDF of polygons that are
+    above the Circular Compaactness threshold.
```
jGaboardi, line 767: *Compactness

momepy/preprocessing.py
```diff
@@ -740,3 +746,343 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+
+    Return
+    ________
+    GeoDataFrames : round abouts and adjacent polygons
```
jGaboardi, line 771: "roundabout" is generally one word.

momepy/preprocessing.py
```diff
@@ -740,3 +746,343 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+    # selecting round about polygons based on compactness
+    mask = circom_serie > circom_threshold
+    rab = gdf[mask]
+    # exclude those above the area threshold
+    area_threshold_val = gdf.area.quantile(area_threshold)
+    rab = rab[rab[area_col] < area_threshold_val]
+
+    if include_adjacent is True:
+
+        # calculating some parameters
```
jGaboardi, line 787: We can probably remove this comment.

momepy/preprocessing.py
```diff
@@ -740,3 +746,343 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+
+        area_right = area_col + "_right"
+        area_left = area_col + "_left"
+        area_mask = rab_adj[area_right] >= rab_adj[area_left]
+        rab_adj = rab_adj[area_mask]
+        rab_adj.index.name = "index"
+
+        # adding a hausdorff_distance threshold
+        rab_adj["hdist"] = 0
+        # TODO: (should be a way to verctorize)
```
jGaboardi, line 808: "vectorize"

momepy/preprocessing.py
```diff
@@ -740,3 +746,343 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+    roundabout.
+    """
+    idx_out_many_incoming = []
+    # For each new connection, evaluate COINS and select the group from which the new
+    # line belongs
+    # TODO ideally use the groupby object on line_wkt used earlier
+    for g, x in incoming_many.groupby("line_wkt"):
+        gs = gpd.GeoSeries(pd.concat([x.geometry, x.line]), crs=incoming_many.crs)
+        gdf = gpd.GeoDataFrame(geometry=gs)
+        gdf.drop_duplicates(inplace=True)
```
jGaboardi, line 886: Sometime its better to use ``gdf = gdf.drop_duplicates()`` for readability, though I personally don't have a strong preference. @martinfleis What are your thoughts on this?

momepy/preprocessing.py
```diff
@@ -740,3 +746,343 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+        edges_idx, rabs_idx = rab_multipolygons.sindex.query_bulk(
+            edges.geometry, op="covered_by"
+        )
+    idx_drop = edges.index.take(edges_idx)
+    touching_idx = touching.index
+    ls = list(set(touching_idx) - set(idx_drop))
+
+    incoming = touching.loc[ls]
+
+    # figuring out which ends of incoming edges needs to be connected to the center_pt
```
jGaboardi, line 925: *need

momepy/preprocessing.py
```diff
@@ -740,3 +746,343 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+    incoming["first_pt"] = incoming.geometry.apply(lambda x: Point(x.coords[0]))
+    incoming["dist_fisrt_pt"] = incoming.center_pt.distance(incoming.first_pt)
```
jGaboardi, line 927: Correct spelling and column names to `"dist_first_pt"`. See also line 932.

momepy/preprocessing.py
```diff
@@ -740,3 +746,343 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+    incoming["dist_last_pt"] = incoming.center_pt.distance(incoming.last_pt)
+    lines = []
+    for i, row in incoming.iterrows():
+        if row.dist_fisrt_pt < row.dist_last_pt:
+            lines.append(LineString([row.first_pt, row.center_pt]))
+        else:
+            lines.append(LineString([row.last_pt, row.center_pt]))
+    incoming["line"] = gpd.GeoSeries(lines, index=incoming.index, crs=edges.crs)
+
+    # checking in there are more than one incoming lines arriving to the same point
```
jGaboardi, line 938: checking *if

momepy/preprocessing.py
```diff
@@ -740,3 +746,343 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+    incoming_all = gpd.GeoDataFrame(
+        pd.concat([incoming_ones, incoming_many_reduced]), crs=edges.crs
+    )
+
+    return incoming_all, idx_drop
+
+
+def _ext_lines_to_center(edges, incoming_all, idx_out):
+    """
+    Extends the Linestrings geometries to the centerpoint defined by
```
jGaboardi, line 966: Best to either stick with 'linestrings' or LineStrings.

momepy/preprocessing.py
```diff
@@ -740,3 +746,343 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+    _rabs_center_points. Also deletes the lines that originally defined the roundabout.
+    Creates a new column labled with the 'rab' number.
+
+    Returns
+    -------
+    GeoDataFrame
+        GeoDataFrame of with updated geometry
```
jGaboardi, line 973: ~~of~~

momepy/preprocessing.py
```diff
@@ -740,3 +746,343 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+    )
+
+    # deleting the original round about edges
+    new_edges = edges.drop(idx_out, axis=0)
+
+    # creating a unique label for returned gdf
+    incoming_label = "rab_" + incoming_all.index_right.astype(str)
+    incoming_label = incoming_label[~incoming_label.index.duplicated(keep="first")]
+
+    # mantianing the same gdf shape that the original
```
jGaboardi, line 987: *maintaining the same gdf shape as the original

momepy/preprocessing.py
```diff
@@ -740,3 +746,343 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+    incoming_all = incoming_all[edges.columns]
+    new_edges = pd.concat([new_edges, incoming_all])
+
+    # adding a new column to match
+    new_edges["simpl_edge"] = np.nan
+    new_edges["simpl_edge"] = incoming_label
```
jGaboardi, line 993: Is line 992 needed or can we simply go directly with 993?

momepy/preprocessing.py
```diff
@@ -740,3 +746,343 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+        - 'mean': calculates the mean coordinates from the points of polygons (including
+         adjacent geometries)
+    angle_threshold : int, float (default 0)
+        The angle threshold for the COINS algorithm. Only used when multiple incoming
+        LineStrings
+        arrive at the same Point to the roundabout or to the adjacent polygons if set
+        as True.
+        eg. when two 'edges' touch the roundabout at the same point, COINS algorithm
+        will evaluate which of those
+        incoming lines should be extended accordinf to their deflection angle.
```
jGaboardi, line 1059: *according

momepy/preprocessing.py
gregmaya: There might be a way to keep this on a single line but because `incoming_label` only has certain matching values to `new_edges`, I was having an error for different lengths... 
@martinfleis can you think of a workaround?

momepy/preprocessing.py
gregmaya: oh true ! ... 
I believe I've seen this notation `gdf = gdf.drop_duplicates()` on the code so I'll stick to that, easy fix!

tests/test_preprocess.py
gregmaya: That makes sense... I'll implement that, thanks !

momepy/preprocessing.py
martinfleis: I belie that if you delete just `new_edges["simpl_edge"] = np.nan` the result will be the same. Pandas matches the series based on the index and fills non-matching observations with NaN itself.

gregmaya:
@martinfleis I've combined all the suggestions by @jGaboardi as well as included a diameter_factor for including adjacent polygons that stretch a bit more than what was considered before. 
I cannot figure out what the error is while trying to build the _Read the Docs_ so a hint there would be nice to potentially merge this PR?! :)

jGaboardi:
@gregmaya We are also seeing the *Read The Docs* error in #382, so it's not specific to this PR. However, I am wondering what is going on with the 3.8 and 3.10-dev tests.

martinfleis:
3.8-minimal has old geopandas. That needs to be catched and we should raise an error saying that the function requires newer version than 0.x.0 (need to check which is min).

3.10 is some issue with networkx main, ignore it here.

gregmaya:
> 3.8-minimal has old geopandas. That needs to be catched and we should raise an error saying that the function requires newer version than 0.x.0 (need to check which is min).
> 
> 3.10 is some issue with networkx main, ignore it here.

.... So drumroll 🥁...?! 🙏😅

martinfleis:
by "that needs to be catched" I meant that you need to catch it and raise a warning :)

momepy/preprocessing.py
```diff
@@ -1067,6 +1068,14 @@ def roundabout_simplification(
     GeoDataFrame
         GeoDataFrame with an updated geometry and an additional column labeling modified edges.
     """
+    ver_str = gpd.__version__
+    ver = re.search("([\d.]).([\d.]+).([\d]+)", ver_str).group(2)
```
jGaboardi, line 1072: This method for determining version compatibility seems a bit cumbersome and would produce desired results for version over `v1.0.0` for example. Let try something simpler, like which is already in [here](https://github.com/pysal/momepy/blob/main/momepy/preprocessing.py#L28).

gregmaya: that's much cleaner, indeed... 
(slow learnings, here)

jGaboardi: Don't be hard on yourself! Everything takes some practice!

jGaboardi: We'll either need to skip tests for that function if version is lower than 0.9.0 ***or*** test that the error is caught if using those versions. The second choice is probably what we want to do. [Here's an example](https://github.com/pysal/momepy/blob/main/tests/test_graph.py#L42) of how to do that.

momepy/preprocessing.py
```diff
@@ -740,3 +746,352 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+    -------
+    GeoDataFrame
+        GeoDataFrame with an updated geometry and an additional column labeling modified edges.
+    """
+    ver_str = gpd.__version__
+    if Version(ver_str) <= Version("0.9"):
+        raise TypeError(
+            "The function needs at least GeoPandas 0.9.0. Your current version is "
+            + ver_str
+        )
```
martinfleis, line 1075: On line 34, there is

```py
GPD_10 = Version(gpd.__version__) >= Version("0.10")
```

Add 
```py
GPD_09 = Version(gpd.__version__) >= Version("0.9")
```
there and then use this here:

```suggestion
    if not GPD_09:
        raise ImportError(
            f"`roundabout_simplification` requires geopandas 0.9.0 or newer. Your current version is {gpd.__version__}."
        )
```

momepy/preprocessing.py
gregmaya: @jGaboardi thanks for that example... I admit I'm not familiar with the WITH statement and therefore will need to get my head around to be able to skip all tests if GDP version is less than 0.9.0 (which by the way is when 'predicate `touches` seems to have been surfaced according to the documentation of 0.8.1 and 0.8.2)

jGaboardi: So maybe we should add the conditional for less than `0.8.1`?

As for the `with` statement, it is used to isolate a specific block of code. For use in testing it is nice because we can run a test that we know will fail, for example [here](https://github.com/pysal/momepy/blob/main/tests/test_utils.py#L32-L34) we know that a `ValueError` should be raised and we can test for it. In our case, we know that an `ImportError` should be raised and we can test for that *only if* we don't have the proper version of GeoPandas (made `0.8.1`?).

gregmaya: sorry what I meant to say is that the first time the predicate 'touches' appears is on 0.9.0 (`touches` is used in a couple of places during the process - so, actually, quite vital for overall success)
[Here are the docs of 0.8.2 ](https://geopandas.org/en/v0.8.2/reference/geopandas.sjoin.html?highlight=sjoin#geopandas.sjoin) for reference.

as for the `with` statement, I think I understand the idea behind but I just don't really know what should be tested (if anything at all) when geopandas is less than 0.9.0... 

my current line 139 : 
`with pytest.raises(TypeError): # unsure how to develop this`

jGaboardi: Ah, OK. The purpose is to ensure errors are raised when we know they should be raised. And since we know that we need at least version 0.9.0 to use the roundabouts simplification, we should test that an `ImportError` is raised if using less than that (because we added that stipulation into `roundabout_simplification()`. So maybe something the following would be good and then repeat for each of the 5 simplification tests that you wrote:

```python
def test_roundabout_simplification_1(self):
     if not GPD_09:
         with pytest.raises(ImportError):
             mm.roundabout_simplification(self.df_streets_rabs)
     else:
         check = mm.roundabout_simplification(self.df_streets_rabs)
         assert len(check) == 65
         assert len(self.df_streets_rabs) == 88  # checking that nothing has changed
```

And along the same lines, we can also check that the `TypeError` is raised properly by creating a test that passing in MultiLineStrings, Points, etc...

gregmaya: OK yeah, that makes more sense in my head now! 
I've used your code for test_1 but I only create an if clause that skips the tests when GPD_09 is false (i.e. less than 0.9.0) for the other 4 tests. 

As for the TypeError, do you mean that it would be worth adding an extra layer to the file current _gpkg_ with Multi geometries and another one with Points to catch the errors ?! -> just confirming before I go ahead and change files and stuff. :) 
I appreciate the extra attention this is taking, btw!

martinfleis: > As for the TypeError, do you mean that it would be worth adding an extra layer to the file current gpkg with Multi geometries and another one with Points to catch the errors ?! -> just confirming before I go ahead and change files and stuff. :)

Not worth it.

tests/test_preprocess.py
```diff
@@ -124,3 +132,44 @@ def test_extend_lines(self):
         ext5 = mm.extend_lines(gdf, 2)
         assert ext5.length.sum() > gdf.length.sum()
         assert ext5.length.sum() == pytest.approx(6.2, rel=1e-3)
+
+    def test_roundabout_simplification_1(self):
```
martinfleis, line 136: Can you rename these and instead of `_1` `_2` `_3` use something like `_default`, `_high_threshold`, `_low_threshold`, so we know what aspect is the test checking from the name?

tests/test_preprocess.py
```diff
@@ -124,3 +132,44 @@ def test_extend_lines(self):
...
+            assert len(check) == 65
+            assert len(self.df_streets_rabs) == 88  # checking that nothing has changed
+
+    def test_roundabout_simplification_2(self):
+        if GPD_09:
+            check = mm.roundabout_simplification(
+                self.df_streets_rabs, self.df_rab_polys, circom_threshold=0.97
+            )
+            assert len(check) == 77
+            assert len(self.df_streets_rabs) == 88
```
martinfleis, line 151:
```suggestion
    @pytest.mark.skipif(not GPD_09, reason="requires geopandas 0.9+")
    def test_roundabout_simplification_2(self):
        check = mm.roundabout_simplification(
            self.df_streets_rabs, self.df_rab_polys, circom_threshold=0.97
        )
        assert len(check) == 77
        assert len(self.df_streets_rabs) == 88
```

This is another trick pytest can do. Instead of condition you use this decorator and the test is skipped for older geopandas. Use the same for all of them.

tests/test_preprocess.py
```diff
@@ -124,3 +132,44 @@ def test_extend_lines(self):
         ext5 = mm.extend_lines(gdf, 2)
         assert ext5.length.sum() > gdf.length.sum()
         assert ext5.length.sum() == pytest.approx(6.2, rel=1e-3)
+
+    def test_roundabout_simplification_1(self):
+        if not GPD_09:
+            with pytest.raises(ImportError):
+                mm.roundabout_simplification(self.df_streets_rabs)
```
martinfleis, line 139: See the comment on decorators below. What I would do is to turn this into a test by itself.

```py
@pytest.mark.skipif(GPD_09, reason="requires geopandas 0.9+")
def test_roundabout_simplification_gpd_error(self):
    with pytest.raises(ImportError, match="`roundabout_simplification` requires geopandas 0.9.0"):
        mm.roundabout_simplification(self.df_streets_rabs)
```

This is to make sure each test focuses on one thing to check, in this case an error. I am also adding `match` to check that the exact error is raised, not just some ImportError.

martinfleis (commented):
Some minor bits on test. Then we're ready to go.

tests/test_preprocess.py
```diff
@@ -124,3 +132,48 @@ def test_extend_lines(self):
...
         assert ext5.length.sum() == pytest.approx(6.2, rel=1e-3)
+
+    @pytest.mark.skipif(GPD_09, reason="requires geopandas 0.9+")
+    def test_roundabout_simplification_gpd_error(self):
+        with pytest.raises(
+            ImportError, match="`roundabout_simplification` requires geopandas 0.9.0"
+        ):
+            mm.roundabout_simplification(self.df_streets_rabs)
+
+    @pytest.mark.skipif(GPD_09, reason="requires geopandas 0.9+")
```
martinfleis, line 143:
```suggestion
    @pytest.mark.skipif(not GPD_09, reason="requires geopandas 0.9+")
```

jGaboardi (approved):

momepy/preprocessing.py
```diff
@@ -740,3 +747,350 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+    # creating a unique label for returned gdf
+    incoming_label = "rab_" + incoming_all.index_right.astype(str)
+    incoming_label = incoming_label[~incoming_label.index.duplicated(keep="first")]
+
+    # maintaining the same gdf shape as the original
+    incoming_all = incoming_all[edges.columns]
+    new_edges = pd.concat([new_edges, incoming_all])
+
+    # adding a new column to match
+    new_edges["simpl_edge"] = incoming_label
```
martinfleis, line 989: `"simpl_edge"` is not a very intuitive name. Same as the value that comes with it (i.e. `rab_12`). It should be clear what it is and what it refers to. What is `12` in `rab_12`? Why do I get 3 groups if I use your testing data that are `rab_6`, `rab_8` and `rab_12`? Where does this come from in relation to my original data? Why do I get it back?

gregmaya: Let me explain further to come up with better namings.
`simple_edge` --> short for 'simplified_edge' : This is a column that's aimed at identifying the edges that have been altered during this process. In my head, the upcoming methods() will further populate this same column. 

`rab_` --> short for 'roundabout'. The number comes originally from the `id` of the roundabout polygon which is used to identify the overall group (including adjacent polygons when appropriate).

martinfleis: Okay. I think that if we're adding this as a column to the data that didn't have it previously, it needs to be useful for the user. I would call the column `simplification_group` and try to return numerical ids, not strings with rab_. Why are some values missing? I would expect I'd get 1, 2, 3, not 6, 8, 12.

gregmaya: just made said changes ...

martinfleis (approved):
🎉

jGaboardi:
congrats @gregmaya!!! :tada:

PR #377, opened 2022-08-09
ENH: improved node consolidation functions

gsagostini:
Function for consolidating intersections on street networks based on graph-distances along with edge reconstruction. Tests to come.


momepy/preprocessing.py
```diff
@@ -15,6 +15,11 @@
 from packaging.version import Version
 from tqdm.auto import tqdm
 
+import networkx as nx
```
jGaboardi, line 18: Let's run `black` on this PR.

gsagostini: new commit with black format

momepy/preprocessing.py
```diff
@@ -740,3 +745,368 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+
+    tolerance : float
+        distance in network units below which nodes will be consolidated
+
+    rebuild_graph: Boolean
+
+    rebuild_edges_method: string
+        'spider' or 'linear' or 'crow_flies'
+
+    directed: Boolean
```
martinfleis, line 797: In which case would you not infer it from the graph?

momepy/preprocessing.py
```diff
@@ -740,3 +745,368 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+    ----------
+    graph : Networkx.MultiGraph or Networkx.MultiDiGraph
+
+    tolerance : float
+        distance in network units below which nodes will be consolidated
+
+    rebuild_graph: Boolean
+
+    rebuild_edges_method: string
+        'spider' or 'linear' or 'crow_flies'
```
martinfleis, line 795: I am not sure if these options are clear based on their names. `spider` is okay but as crow flies is linear line... What about using `extend` instead of `linear` and `euclidean` instead of `crow_flies`?

momepy/preprocessing.py
```diff
@@ -740,3 +745,368 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+    Graph must be configured so that
+
+    1. All nodes have attributes determining their x and y coordinates;
+    2. All edges have attributes determining their origin, destination, and geometry;
+    3. Graph itself has an attribute determining its crs.
+
+    Parameters
+    ----------
+    graph : Networkx.MultiGraph or Networkx.MultiDiGraph
+
```
martinfleis, line 788: remove empty lines between parameters

momepy/preprocessing.py
```diff
@@ -740,3 +745,368 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+    """
+
+    # Collect nodes and their data:
+    nodes, nodes_dict = zip(*graph.nodes(data=True))
+    nodes_df = pd.DataFrame(nodes_dict, index=nodes)
+    nodes_gdf = gpd.GeoDataFrame(
+        nodes_df,
+        crs=graph.graph["crs"],
+        geometry=gpd.points_from_xy(nodes_df[x_col], nodes_df[y_col]),
+    )  # improve crs here
```
martinfleis, line 818: what do you mean by this comment?

momepy/preprocessing.py
```diff
@@ -740,3 +745,368 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+        [
+            edge
+            for edge in graph.edges(keys=True, data=True)
+            if edge[-1]["length"] > tolerance
+        ]
+    )
+    isolated_nodes_list = list(nx.isolates(components_graph))
+    components_graph.remove_nodes_from(isolated_nodes_list)
+
+    # The connected components of this graph are node clusters we must individually simplify. We collect them in a dataframe and
```
martinfleis, line 836: try to make sure that all lines, including comments, are shorter than 88 characters. I think that `flake8` will complain here, though we don't have it on in CI. (@jGaboardi maybe we should? also `isort`?)

momepy/preprocessing.py
```diff
@@ -740,3 +745,368 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+        ],
+        axis=1,
+        join="inner",
+    )
+
+    # The two node attributes we need for the clusters are the position of the cluster centroids. Those are obtained by averaging
+    #  the x and y columns. We also add attribtues referring to the original osmids in every cluster:
+    cluster_centroids_df = nodes_to_merge_df.groupby("cluster").mean()
+    cluster_centroids_df["simplified"] = True
+    cluster_centroids_df["original_osmids"] = cluster_centroids_df.index.map(
```
martinfleis, line 859: This assumes `osmnx`-like graph, doesn't it? At least in the name.

momepy/preprocessing.py
```diff
@@ -740,3 +745,368 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+    edge_from_col="from",
+    edge_to_col="to",
+):
+    """
+    Rebuilds edge given (potential) new origin and destinations through a linear method: current line is linearly extended
+     to and from the new endpoints.
+
+    :param geometry: shapely.LineString, current edge
+    :param new_origin: shapely.Point or None, new origin point or None if origin remains the same
+    :param new_destination: shapely.Point or None, new destination point or None if destination remains the same
```
martinfleis, line 906: Since you document this, can you follow the style above? For the sake of consistency.

momepy/preprocessing.py
```diff
@@ -740,3 +745,368 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+        lambda v: nodes_dict[v] if v in nodes_dict else -1
+    )
+
+    # Determine what edges need to be simplified (either between diff. clusters or self-loops in a cluster):
+    edges_gdf.query(
+        "origin_cluster != destination_cluster or (('"
+        + edge_to_col
+        + "' == '"
+        + edge_from_col
+        + "') and origin_cluster >= 0)",
```
martinfleis, line 924: This is probably just a personal preference, but could you make this an f-string instead of this?

momepy/preprocessing.py
```diff
@@ -740,3 +745,368 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+        inplace=True,
```
martinfleis, line 925: Another preference - can you not use `inplace=True` but rather return a new object? It is easier to track what is happening. At least that is what we tend to do in pandas world. I realise that networkx model is a bit different with all methods working in place (hence needing `deepcopy`)

momepy/preprocessing.py
```diff
@@ -740,3 +745,368 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+                edge.geometry,
+                edge.new_origin_pt,
+                edge.new_destination_pt,
+                1.5 * tolerance,
+            ),
+            axis=1,
+        )
+
+    # Rename and update the columns:
+    edges_gdf.reset_index(inplace=True)
```
martinfleis, line 964: Again, try not ti use `inplace`.

martinfleis: Same below a couple of times.

momepy/preprocessing.py
```diff
@@ -740,3 +745,368 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+            "origin_cluster": edge_from_col,
+            "destination_cluster": edge_to_col,
+            "geometry": "original_geometry",
+            "new_geometry": "geometry",
+        },
+        axis=1,
+        inplace=True,
+    )
+    edges_gdf.drop(columns=["new_origin_pt", "new_destination_pt"], inplace=True)
+    edges_gdf["length"] = edges_gdf["geometry"].apply(lambda x: x.length)
```
martinfleis, line 978:
```suggestion
    edges_gdf["length"] = edges_gdf.length
```
This will use vectorised operation instead of a loop.

martinfleis: Btw, why do you need this anyway?

momepy/preprocessing.py
```diff
@@ -740,3 +745,368 @@ def _get_extrapolated_line(coords, tolerance, point=False):
...
+        current_node = Point(line_coords[0])
+        geometry = linemerge([LineString([new_origin, current_node]), geometry])
+    # Assuming the line is not closed, we can find its endpoints via the boundary attribute:
+    else:
+        current_origin, current_destination = geometry.boundary
+        if new_origin is not None:
+            # Create a buffer around the new origin:
+            new_origin_buffer = new_origin.buffer(buff)
+            # Use shapely.ops.split to break the edge where it intersects the buffer:
+            geometry_split_by_buffer_list = list(split(geometry, new_origin_buffer))
```
martinfleis, line 1049: have you tried using `difference` instead? It may be more efficient that `split` + `linemerge` (but may not and may cause more trouble than good).

momepy/preprocessing.py
gsagostini: Potentially user has a MultiDiGraph but wants to treat it as MultiGraph so inferring works against it. This is the case with OSMnx graphs if one doesn't run "get_undirected" first. If you think it's more straightforward to always infer, will remove the param.

martinfleis: Keep it there, just add something like you just wrote as an explanation.

momepy/preprocessing.py
gsagostini: maybe add support to graphs which don't have crs attribute. currently, function fails in such case.

momepy/preprocessing.py
gsagostini: good catch. is "original_nodes" a better name?

momepy/preprocessing.py
martinfleis: Makes sense. Something like `crs=graph.graph.get("crs")`?

momepy/preprocessing.py
martinfleis: Yeah

momepy/preprocessing.py
gsagostini: Does that return None if key is not in dictionary? And do you know whether I can build a GeoDataFrame with crs None?

momepy/preprocessing.py
gsagostini: Updates length parameter of rebuilt edges to match their new geometries.

momepy/preprocessing.py
martinfleis: Yes and yes.

momepy/preprocessing.py
jGaboardi: Seems to be some good options for [`flake8`](https://github.com/py-actions/flake8) and [`isort`](https://github.com/isort/isort-action) with GHA. Would not hurt at all to give it a try. Also, [`isort`+`black` compatibility with `pre-commit`](https://pycqa.github.io/isort/docs/configuration/black_compatibility.html).

martinfleis (commented):
@gsagostini have you had a chance to work on tests?

martinfleis:
@gsagostini Have you closed it intentionally?

gsagostini:
I think I did it unintentionally while trying to fetch upstream in order to bring my fork up to date---but it may be easier to open a new one at this point, would that be bad?

gsagostini:
@martinfleis @jGaboardi I reopened it with the tests and updated commits.

jGaboardi:
@gsagostini Did you run `black` with `pre-commit` for code in this PR before pushing up the latest commits?

gsagostini:
@jGaboardi I had ran black only in the momepy directory but now it should have been fine for the tests too

martinfleis:
You will need to add `consolidate_intersections` to the list called `__all__` on top of the file.

gsagostini:
Tests are still failing for the `consolidate_intersections` function on the minimal environment. Is this expected? Seems like the tests fail for other functions too, not just for the one I added

tests/test_preprocess.py
```diff
@@ -27,6 +28,8 @@ def setup_method(self):
         self.df_rab_polys = gpd.GeoDataFrame(
             geometry=[g for g in plgns], crs=self.df_streets_rabs.crs
         )
+        test_file_path3 = mm.datasets.get_path("nyc_graph", extension="graphml")
+        self.graph = ox.get_undirected(ox.load_graphml(test_file_path3))
```
jGaboardi, line 32: Not sure exactly what is going on, but it appears to be an issue with `ox.load_graphml()`. Can you confirm that the file path in `test_file_path3` is correct?

gsagostini: The file path is corrected according to the rewriting of the get_path function to deal with other extensions. Is it possible that osmnx is not included in the minimal test environment?

jGaboardi: hmmm. `osmnx` is in the [minimal](https://github.com/pysal/momepy/blob/afcb85b89195f3ba2c5c869fb2e5f5add357c927/ci/envs/38-minimal.yaml#L14) testing environment.

jGaboardi: But it does seems likely that some dependency is missing in that environment and is causing the error. Or a dependency that is needed is not supported by Python 3.8. If we look [here](https://github.com/gboeing/osmnx/blob/main/osmnx/io.py#L252) within `osmnx`, the `.read_graphml()` is being called from `networkx` with keyword `force_multigraph=True`, which we can see in the CI error log [here](https://github.com/pysal/momepy/runs/8294778205?check_suite_focus=true#step:6:1289).

jGaboardi: So after some more digging we can see that the version of `networkx` being pulled in for the Python 3.8 test is [v2.3](https://github.com/pysal/momepy/runs/8294778205?check_suite_focus=true#step:5:157), which was [released in 2019](https://github.com/networkx/networkx/releases/tag/networkx-2.3). And that version of `networkx` did not support the `force_multigraph` keyword as in does now (see [v.2.3 `GraphMLReader`](https://github.com/networkx/networkx/blob/807d9fde187e18a58d542f87ab0d0cb98ef46459/networkx/readwrite/graphml.py#L237) vs. [current `main`'s `GraphMLReader`](https://github.com/networkx/networkx/blob/034274e08b3b78022e47d03cc5e69e28be23f713/networkx/readwrite/graphml.py#L293)).

So the trick now is to find the earliest version of `networkx` that supports `force_multigraph` in the `GraphMLReader` signature and [update the pin in the CI environment](https://github.com/pysal/momepy/blob/afcb85b89195f3ba2c5c869fb2e5f5add357c927/ci/envs/38-minimal.yaml#L8).

jGaboardi: And it looks like that will be [`networkx=2.5`](https://github.com/networkx/networkx/blob/12f852f6a686ed32f718125c5965586dd6bb82a9/networkx/readwrite/graphml.py#L199).

gsagostini: Thanks a lot, James! I wouldn't have been able to figure this out. Updated it and let's see if it goes through now

momepy/preprocessing.py
```diff
@@ -1101,3 +1105,453 @@ def roundabout_simplification(
...
+    method: string
+        'extension' or 'spider' or 'euclidean'
+    buffer : float
+        distance to buffer consolidated nodes in the Spider-web reconstruction
+    edge_from_col, edge_to_col: string
+        edge attribute with the valid origin/destination node id
+
+    Returns
+    ----------
+    List
```
jGaboardi, line 1296: What's the name of this variable?

momepy/preprocessing.py
```diff
@@ -1101,3 +1105,453 @@ def roundabout_simplification(
...
+    nodes_dict: dict
+        Dictionary whose keys are node ids and values are the corresponding consolidated
+        node cluster ids. Only consolidated nodes are in the dictionary.
+    cluster_gdf : GeoDataFrame
+        GeoDataFrame containing consolidated node ids.
+    method: string
+        'extension' or 'spider' or 'euclidean'
+    buffer : float
+        distance to buffer consolidated nodes in the Spider-web reconstruction
+    edge_from_col, edge_to_col: string
```
jGaboardi, line 1291: Add these keywords on separate lines.

momepy/preprocessing.py
```diff
@@ -1101,3 +1105,453 @@ def roundabout_simplification(
...
+        consider the graph a MultiDiGraph if True or MultiGraph if False, and
+        if None infer from the passed object type
+    x_col, y_col: string
+        node attribute with the valid coordinate
+    edge_from_col, edge_to_col: string
+        edge attribute with the valid origin/destination node id
+
+    Returns
+    ----------
+    Networkx.MultiGraph or Networkx.MultiDiGraph
```
jGaboardi, line 1162: What's the name of this variable?

momepy/preprocessing.py
```diff
@@ -1101,3 +1105,453 @@ def roundabout_simplification(
...
+        distance in network units below which nodes will be consolidated
+    rebuild_graph: Boolean
+    rebuild_edges_method: string
+        'extend' or 'spider' or 'euclidean'
+    directed: Boolean or None
+        consider the graph a MultiDiGraph if True or MultiGraph if False, and
+        if None infer from the passed object type
+    x_col, y_col: string
+        node attribute with the valid coordinate
+    edge_from_col, edge_to_col: string
```
jGaboardi, line 1157: Keyword arguments should be described on individual lines for docstrings.

momepy/preprocessing.py
```diff
@@ -1101,3 +1105,453 @@ def roundabout_simplification(
...
+            lambda edge: _spider_simplification(
+                edge.geometry, edge.new_origin_pt, edge.new_destination_pt, buffer
+            ),
+            axis=1,
+        )
+        edges_simplified_gdf = edges_tosimplify_gdf.assign(
+            new_geometry=edges_simplified_geometries
+        )
+    else:
+        print("Simplification method not recognized. Using spider-web simplification.")
```
jGaboardi, line 1359: Should this throw a `warning` rather than a `print` statement?

momepy/preprocessing.py
```diff
@@ -1101,3 +1105,453 @@ def roundabout_simplification(
...
+    """
+    Extends edge geometry to new endpoints.
+
+    If either new_origin or new_destination is None, maintains the
+      respective current endpoint.
+
+    Parameters
+    ----------
+    geometry : shapely.LineString
+    new_origin, new_destination: shapely.Point or None
```
jGaboardi, line 1415: Should be two entries. But Since this is a non-user facing function maybe it doesn't need detailed docstrings? Maybe type hints though? @martinfleis What are your thoughts on this?

momepy/preprocessing.py
```diff
@@ -1101,3 +1105,453 @@ def roundabout_simplification(
...
+
+    Returns
+    ----------
+    shapely.LineString
```
jGaboardi, line 1419: report the variable name and description if we decide for a detailed docstring here.

momepy/preprocessing.py
```diff
@@ -1101,3 +1105,453 @@ def roundabout_simplification(
...
+      current geometry within a buffer of the new endpoint and then extends
+      it linearly. Useful to maintain planarity.
+
+    If either new_origin or new_destination is None, maintains the
+      respective current endpoint.
+
+    Parameters
+    ----------
+    geometry : shapely.LineString
+    new_origin, new_destination: shapely.Point or None
```
jGaboardi, line 1450: Same as above.

momepy/preprocessing.py
```diff
@@ -1101,3 +1105,453 @@ def roundabout_simplification(
...
+    buff : float
+        distance from new endpoint to break current geometry
+
+    Returns
+    ----------
+    shapely.LineString
```
jGaboardi, line 1456: Same as above.

momepy/preprocessing.py
```diff
@@ -1101,3 +1105,453 @@ def roundabout_simplification(
...
+    Rebuilds edge geometry to new endpoints. Ignores current geometry
+      and traces a straight line between new endpoints.
+
+    If either new_origin or new_destination is None, maintains the
+      respective current endpoint.
+
+    Parameters
+    ----------
+    geometry : shapely.LineString
+    new_origin, new_destination: shapely.Point or None
```
jGaboardi, line 1536: Same as above.

momepy/preprocessing.py
```diff
@@ -1101,3 +1105,453 @@ def roundabout_simplification(
...
+
+    Returns
+    ----------
+    shapely.LineString
```
jGaboardi, line 1540: Same as above.

momepy/preprocessing.py
gsagostini: will replace that. also noticed such lines aren't covered by tests so will add a new test case
```suggestion
        import warnings
        warnings.warn("Simplification method not recognized. Using spider-web simplification.")
```

momepy/preprocessing.py
martinfleis: That doesn't really matter, does it? It may be not even be assigned to any variable.

momepy/preprocessing.py
gsagostini: for the sake of consistency I fixed in all cases

momepy/preprocessing.py
```diff
@@ -1101,3 +1105,453 @@ def roundabout_simplification(
...
+    method="spider",
+    buffer=45,
+    edge_from_col="from",
+    edge_to_col="to",
+):
+    """
+    Update origin and destination on network edges when original endpoints were replaced by a
+      consolidated node cluster. New edges are drawn according to method which is one of:
+
+    1. Extension reconstruction:
```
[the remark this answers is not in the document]
gsagostini: I would argue to leave them in the docstring anyways a the `_get_rebuilt_edges` functions is not user facing

momepy/preprocessing.py
gsagostini: I think I did not properly name output variables. I usually include in the function description a brief explanation of the output and then just give the type below the Returns block. Should I arbitrarily name it based on what it represents?

momepy/preprocessing.py
jGaboardi: 1. We also might want to clean the `method` variable as `method = method.lower()` to correct for users accidentally capitalizing, etc.
2. We can make this more user-friendly by reporting the method passed in. Something like: ``warnings.warn("Simplification method (`'{method}'`) not recognized. Using spider-web simplification.")``

momepy/preprocessing.py
martinfleis: I don't think so, it is fine as it is imho. That applies to all cases of this.

jGaboardi: My personal preference is explicit variable naming for descriptive docstrings, but I will defer the @martinfleis's decision on this.

momepy/preprocessing.py
```diff
@@ -1101,3 +1105,464 @@ def roundabout_simplification(
...
+            ),
+            axis=1,
+        )
+        edges_simplified_gdf = edges_tosimplify_gdf.assign(
+            new_geometry=edges_simplified_geometries
+        )
+    else:
+        warnings.warn(
+            f"Simplification method '{method}' not recognized. Using spider-web simplification."
+        )
```
martinfleis, line 1367: I don't think this is a good behaviour. We should raise an error if the method is not implemented.

momepy/preprocessing.py
```diff
@@ -1101,3 +1105,464 @@ def roundabout_simplification(
...
+    cols_rename = {
+        edge_from_col: "original_from",
+        edge_to_col: "original_to",
+        "origin_cluster": edge_from_col,
+        "destination_cluster": edge_to_col,
+        "geometry": "original_geometry",
+        "new_geometry": "geometry",
+    }
+    cols_drop = ["new_origin_pt", "new_destination_pt"]
+    new_edges_gdf = edges_simplified_gdf.rename(cols_rename, axis=1).drop(
```
martinfleis, line 1388: I am not entirely sure how geopandas handles this specific renaming of geometry columns but I would rather throw in `set_geometry('geometry')` just to be sure the right column is set as active.

martinfleis (commented):
Thanks! This is a great PR! Left just a couple of minor comments.

momepy/preprocessing.py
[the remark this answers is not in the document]
jGaboardi: Fair enough.

momepy/preprocessing.py
```diff
@@ -1101,3 +1105,464 @@ def roundabout_simplification(
...
+    graph : Networkx.MultiGraph or Networkx.MultiDiGraph
+    tolerance : float
+        distance in network units below which nodes will be consolidated
+    rebuild_graph: Boolean
+    rebuild_edges_method: string
+        'extend' or 'spider' or 'euclidean', ignored if rebuild_graph is False
+    directed: Boolean or None
+        consider the graph a MultiDiGraph if True or MultiGraph if False, and
+        if None infer from the passed object type
+    x_col: string
```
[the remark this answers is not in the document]
gsagostini: right, I thought about it as from the gdf. This is the name of the attribute containing the x-coordinate (so that it would work even if it was named longitude for example). Maybe `x_att` is a good name?

martinfleis:
Hi @gsagostini, I know you are likely super busy these days, but do you have an idea when you'll find some time to finalise this PR? I would like to plan the next release and knowing the timeline would certainly help.

gsagostini:
Hi @martinfleis apologies this totally fell out of my radar. I have a day off tomorrow and will address the conflicts from this PR + push the notebook with documentation by my end of day---worst case, the notebook will be added on Wednesday. Hopefully you can take a look at it over the rest of the week, and then we can merge it by the weekend. Sounds good?

martinfleis:
That sounds great! Thanks!

momepy/preprocessing.py
```diff
@@ -1101,3 +1105,464 @@ def roundabout_simplification(
...
+    )
+    cluster_gdf = gpd.GeoDataFrame(
+        cluster_centroids_df, crs=graph_crs, geometry=cluster_geometries
+    )
+    cluster_nodes_list = list(cluster_gdf.to_dict("index").items())
+
+    # Rebuild edges if necessary:
+    if rebuild_graph:
+        rebuild_edges_method = rebuild_edges_method.lower()
+        graph.graph["approach"] = "primal"
```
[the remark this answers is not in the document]
gsagostini: Thanks! Wasn't my intention and I didn't catch that. Moved the `simplified_graph` assignment above this block so that I add the approach to that object instead

momepy/preprocessing.py
```diff
@@ -1101,3 +1105,464 @@ def roundabout_simplification(
...
+    edges i.e. graph is intentionally disconnected.
+
+    Graph must be configured so that
+
+    1. All nodes have attributes determining their x and y coordinates;
+    2. All edges have attributes determining their origin, destination, and geometry.
+
+    Parameters
+    ----------
+    graph : Networkx.MultiGraph or Networkx.MultiDiGraph
```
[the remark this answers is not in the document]
martinfleis: sounds good to me

momepy/preprocessing.py
gsagostini: threw a `ValueError` in that case

momepy/preprocessing.py
```diff
@@ -1101,3 +1105,464 @@ def roundabout_simplification(
...
+    tolerance : float
+        distance in network units below which nodes will be consolidated
+    rebuild_graph: Boolean
+    rebuild_edges_method: string
+        'extend' or 'spider' or 'euclidean', ignored if rebuild_graph is False
+    directed: Boolean or None
+        consider the graph a MultiDiGraph if True or MultiGraph if False, and
+        if None infer from the passed object type
```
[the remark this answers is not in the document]
gsagostini: sounds good, also added in the documentation for the returned object that directionality is inferred from input

gsagostini:
@martinfleis I'm having trouble bringing my fork up to date to solve conflicts. Is there a better way to do so than to revert my commits, fork momepy again, and re-open the PR? I think doin so may close this PR and open a new one in place.

martinfleis:
@gsagostini have you pushed all your changes? I can resolve the conflicts for you.

jGaboardi:
> @martinfleis I'm having trouble bringing my fork up to date to solve conflicts. Is there a better way to do so than to revert my commits, fork momepy again, and re-open the PR? I think doin so may close this PR and open a new one in place.

@gsagostini No need to do this. As @martinfleis said, we can resolve the conflicts if you have completed all your intended commits.

gsagostini:
@jGaboardi @martinfleis thanks! yes those are all commits I wanted to do in the `preprocessing.py` file. I need to fetch the new user guide structure (with the preprocessing directory) to add the notebook, is it better to do it in a new PR once this one gets merged? Also I think CodeFactor check is not successful, but I can't see what lines of code are causing it---is it an issue?

martinfleis:
@gsagostini I synced this with main. You will need to pull the changes. 

flake8 is complaining about long lines in docstrings, you'll need to fix that. CodeFactor can be ignored.

gsagostini:
@martinfleis sounds good will do. Is flake8 something I can install and run like black?

martinfleis:
Yes -> https://flake8.pycqa.org/en/latest/. We also use `isort` now to check the order of imports. The easiest way of ensuring it is all checked on every commit is to install a pre-commit hook. 

You propably must first install pre-commit:

```
pip install pre-commit
```
From the root of the repository, you should then install the pre-commit included in momepy:

```
pre-commit install
```
Then black and flake8 and sort will be run automatically each time you commit changes. You can skip these checks with `git commit --no-verify`.

gsagostini:
Great! I added the notebook too. Let me know if there is more to be said on the user guide, and I will take a look :)

momepy/preprocessing.py
```diff
@@ -1105,3 +1109,447 @@ def roundabout_simplification(
...
+    new_destination: shapely.Point or None
+
+    Returns
+    ----------
+    shapely.LineString
+
+    """
+    # If we are dealing with a self-loop the line has no endpoints:
+    if new_origin == new_destination:
+        current_node = Point(line_coords[0])
```
jGaboardi, line 1420: Where is `line_coords` coming from?

momepy/preprocessing.py
```diff
@@ -1105,3 +1109,447 @@ def roundabout_simplification(
...
+    Networkx.MultiGraph or Networkx.MultiDiGraph
+        directionality inferred from input type
+
+    """
+    # Collect nodes and their data:
+    nodes, nodes_dict = zip(*graph.nodes(data=True))
+    nodes_df = pd.DataFrame(nodes_dict, index=nodes)
+    nodes_geometries = gpd.points_from_xy(nodes_df[x_att], nodes_df[y_att])
+    graph_crs = graph.graph.get("crs")
+    nodes_gdf = gpd.GeoDataFrame(
```
jGaboardi, line 1175: From [`linting`](https://github.com/pysal/momepy/actions/runs/3237046086/jobs/5303895833#step:4:65) we can see that `nodes_gdf` is declared, but not actually used anywhere.

martinfleis:
@gsagostini It would be great if this can be finished soon, so we can release 0.6. Any updates from your side? Do you have a bit of time or shall we try to resolve remaining issues?

martinfleis:
I have revived this. Resolved conflicts, fixed shapely 2.0 compatibility, fixed some minor lint issues. 

I did not do thorough review of the code hoping that my former self in 2022 did :D. I believe this can be merged now. All this functionality will undergo some testing and momepy 1.0 may or may not include it, depending on the results.

jGaboardi:
> hoping that my former self in 2022 did :D.

Dangerous assumption 🤣

momepy/preprocessing.py
```diff
@@ -1033,6 +1037,457 @@ def roundabout_simplification(
...
+      and traces a straight line between new endpoints.
+
+    If either new_origin or new_destination is None, maintains the
+      respective current endpoint.
+
+    Parameters
+    ----------
+    geometry : shapely.LineString
+    new_origin : shapely.Point or None
+    new_destination: shapely.Point or None
```
jGaboardi, line 1467:
```suggestion
    new_destination : shapely.Point or None
```

jGaboardi (approved):
All suggestions are minor doc adjustments.

jGaboardi:
@martinfleis shall we squash & merge?

martinfleis:
@gsagostini Thanks for all the work that went into this!

jGaboardi:
As Martin said, congrats @gsagostini !!! :tada:

PR #378, opened 2022-08-12
BUG: make coins independent of gdf index

amorfinv:
Fix bug initially reported by @gregmaya [here](https://github.com/pysal/momepy/discussions/361#discussioncomment-3365328). 

The issue is that ```coins``` depends on ```gdfs``` with a unique index. However, this will not always be guaranteed. With the modification,  a unique index is created by ```coins``` that is used to build back ```coins.stroke_attribute()```

There were some failing tests on my end but unrelated to ```coins```.

-Andres

### DIFF
--- a/momepy/coins.py
+++ b/momepy/coins.py
@@ -66,7 +66,7 @@ class COINS:
         self.already_merged = False
 
         # get indices of original gdf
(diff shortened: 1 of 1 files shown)


martinfleis:
Thanks!

PR #379, opened 2022-08-12
Added warning for gdp_to_nx if geometries are not LineStrings

Thaulino:
First of all: Thanks for momepy, a great tool for spatial analysis and processing. 

The gdp_to_nx function is pretty nice to convert data, previous read from a shape file into a geopanda frame, to a network graph. 
But from my point of view it makes only sense if the given data consist of LineStrings only, because a point will be converted to an edge with the same start and endpoint. The endpoints will not inherit any attributes from the original data.

I added a warning to inform the user about the pitfall, if any datum has a geometry != Linestring

-- Changes:
Added check and warning in function
Added minimal test if warning is thrown

kind regards

### DIFF
--- a/momepy/utils.py
+++ b/momepy/utils.py
@@ -59,8 +59,9 @@ def _generate_primal(G, gdf_network, fields, multigraph, oneway_column=None):
     for row in gdf_network.itertuples():
 
         if (not (isinstance(row.geometry, LineString))):
-            warnings.warn(message="Geometry is not of type LineString",
-                          category=RuntimeWarning)
+            warnings.warn(
+                message="Geometry is not of type LineString", category=RuntimeWarning
+            )
 
         first = row.geometry.coords[0]
         last = row.geometry.coords[-1]
--- a/momepy/utils.py
+++ b/momepy/utils.py
@@ -8,8 +8,7 @@ import geopandas as gpd
 import libpysal
 import networkx as nx
 import numpy as np
-from shapely.geometry import Point
-from shapely.geometry import LineString
+from shapely.geometry import Point, LineString
 
 __all__ = [
     "unique_id",
--- a/momepy/utils.py
+++ b/momepy/utils.py
@@ -2,12 +2,14 @@
 # -*- coding: utf-8 -*-
 
 import math
+import warnings
 
 import geopandas as gpd
 import libpysal
 import networkx as nx
 import numpy as np
 from shapely.geometry import Point
+from shapely.geometry import LineString
 
 __all__ = [
     "unique_id",
@@ -55,6 +57,11 @@ def _generate_primal(G, gdf_network, fields, multigraph, oneway_column=None):
     G.graph["approach"] = "primal"
     key = 0
     for row in gdf_network.itertuples():
+
+        if (not (isinstance(row.geometry, LineString))):
+            warnings.warn(message="Geometry is not of type LineString",
+                          category=RuntimeWarning)
+
         first = row.geometry.coords[0]
         last = row.geometry.coords[-1]
 
--- a/momepy/utils.py
+++ b/momepy/utils.py
@@ -55,13 +55,21 @@ def _generate_primal(G, gdf_network, fields, multigraph, oneway_column=None):
     Helper for gdf_to_nx.
     """
     G.graph["approach"] = "primal"
+
+    if not "LineString" in gdf_network.geom_type.unique():
+        warnings.warn(
+            message="The given network does not contain any LineString. This can lead to unexpected behaviour. The intended usage of the conversion function is with networks made of LineStrings only.",
+            category=RuntimeWarning,
+        )
+
+    if len(gdf_network.geom_type.unique()) > 1:
+        warnings.warn(
+            message="The given network consists of multiple geometry types. This can lead to unexpected behaviour. The intended usage of the conversion function is with networks made of LineStrings only.",
+            category=RuntimeWarning,
+        )
+
     key = 0
     for row in gdf_network.itertuples():
-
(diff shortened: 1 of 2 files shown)


momepy/utils.py
```diff
@@ -55,6 +57,12 @@ def _generate_primal(G, gdf_network, fields, multigraph, oneway_column=None):
     G.graph["approach"] = "primal"
     key = 0
     for row in gdf_network.itertuples():
+
+        if (not (isinstance(row.geometry, LineString))):
+            warnings.warn(
+                message="Geometry is not of type LineString", category=RuntimeWarning
+            )
```
martinfleis, line 64: This can get quite noisy if you have more non-LineString geometries in you GeoDataFrame. I would suggest calling `gdf_network.geom_type.unique()` before the for loop and checking other geometry types against that. That way we raise only one warning, not one for every geom.

Also, I am not sure if we shouldn't raise an error in this case, rather that a warning. @jGaboardi thoughts?

Thaulino: I agree. This way, it will be annoying - printing the warning just once is much better. 
My first idea was to throw an error, too. But then i thought that it may break some applications, which are working, yet. 
I was unsure how the backward compatibility policy is.

martinfleis: Fair enough. Better not to be too restrictive. But I'd extend the warning then, saying that the graph will not represent the geodataframe truthfully in this case and people should remove unsupported geometry types.

jGaboardi: > This can get quite noisy if you have more non-LineString geometries in you GeoDataFrame. I would suggest calling gdf_network.geom_type.unique() before the for loop and checking other geometry types against that. That way we raise only one warning, not one for every geom.
>
>Also, I am not sure if we shouldn't raise an error in this case, rather that a warning. @jGaboardi thoughts?

@Thaulino raises a good point, but throwing a warning every time is a bit overkill. I think your suggestion of single warning is a nice compromise. 

Also, how are we currently handling `MultiLineString` objects?

martinfleis: Multi* objects will raise when we try to access coords. Shapely will raise an error there.

Thaulino: i implemented the suggested changes

momepy/utils.py
```diff
@@ -53,8 +55,22 @@ def _generate_primal(G, gdf_network, fields, multigraph, oneway_column=None):
     Helper for gdf_to_nx.
     """
     G.graph["approach"] = "primal"
+
+    if not "LineString" in gdf_network.geom_type.unique():
+        warnings.warn(
+            message="The given network does not contain any LineString. This can lead to unexpected behaviour. The intended usage of the conversion function is with networks made of LineStrings only.",
```
jGaboardi, line 61: Maybe a nitpick, but since these are nearly the message we could declare it as a variable:

```python
msg = (
    "The given network %s. This can lead to unexpected behaviour. The intended usage of the conversion function is with networks made of LineStrings only."
)
if not "LineString" in gdf_network.geom_type.unique():
        warnings.warn(
             message=msg % "does not contain any LineString.",
             category=RuntimeWarning,
        )
if len(gdf_network.geom_type.unique()) > 1:
        warnings.warn(
            message=msg % "consists of multiple geometry types.",
             category=RuntimeWarning,
         )

Comment (Thaulino) [reply momepy/utils.py]: I like the idea of putting the shared part of the message into a shared variable. This way doubled code is reduced.
I adjusted the way things are splitted up a little bit -> the message is splitted up into a common and a special part on a sentence level. From my point of view this way the code readability is improved.

**Comment by Thaulino** [comment]
Looking forward for another review.

**Review by jGaboardi — APPROVED** [review]

File: momepy/utils.py [hunk 3]
```diff
@@ -2,12 +2,14 @@
...
 
 import math
+import warnings
 
 import geopandas as gpd
 import libpysal
 import networkx as nx
 import numpy as np
 from shapely.geometry import Point
+from shapely.geometry import LineString
```
martinfleis, line 12:
```suggestion
from shapely.geometry import Point, LineString
```
more concise

tests/test_utils.py
```diff
@@ -2,8 +2,10 @@
 import networkx
 import numpy as np
 import osmnx as ox
+import warnings
 import pytest
 from shapely.geometry import LineString
+from shapely.geometry import Point
```
martinfleis, line 8:
```suggestion
from shapely.geometry import LineString, Point
```

tests/test_utils.py
```diff
@@ -20,12 +22,28 @@ def setup_method(self):
...
+        )
 
     def test_dataset_missing(self):
         with pytest.raises(ValueError):
             mm.datasets.get_path("sffgkt")
 
     def test_gdf_to_nx(self):
+
+        # nx = mm.gdf_to_nx(self.df_points)
+        with pytest.warns(RuntimeWarning):
```
martinfleis, line 40:
```suggestion
        with pytest.warns(RuntimeWarning, match="The given network does not contain any LineString."):
```

This is a more precise check that we raise the exact warning we expect.

tests/test_utils.py
```diff
@@ -20,12 +22,28 @@ def setup_method(self):
...
+            nx = mm.gdf_to_nx(self.df_points)
+
+        # nx = mm.gdf_to_nx(self.df_points_and_linestring)
+        with pytest.warns(RuntimeWarning):
```
martinfleis, line 44:
```suggestion
        with pytest.warns(RuntimeWarning, match="The given network consists of multiple geometry types."):
```

Thaulino: Nice - did not know that feature of pytest yet, added match for both warnings

PR #382, opened 2022-09-07
REF: remove deprecated pandas append

martinfleis:
Replacing pandas `append` by `concat` to avoid the FutureWarning.


jGaboardi (approved):
~~Looks good after we fix the doc generation issue.~~

Maybe spoke too soon...

martinfleis:
None of the failures is related to this PR. I think we can merge it. The docs issue is caused by the update of sphinx-pydata-theme but I haven't looked into why is it happening and what should be done about it.

Issue #388, opened 2022-09-10
suppress warnings in dimensions.StreetProfile()


jGaboardi:
### Describe the problem

Two warnings are being thrown from within `dimensions.StreetProfile()` that should probably be caught and suppressed. Explanations here and MWEs below.

1. https://github.com/pysal/momepy/blob/main/momepy/dimension.py#L640 throws `RuntimeWarning: invalid value encountered in long_scalars.` when there are no `NaN`s in `s` and all `False` in `f`. This would be a `ZeroDivisionError` outside of `numpy`.
2. https://github.com/pysal/momepy/blob/main/momepy/dimension.py#L641 throws `RuntimeWarning: Degrees of freedom <= 0 for slice.` when there are only `NaN`s in `s`.

* See also https://github.com/martinfleis/sdss22-workshop/issues/8

### Steps to reproduce

1. MWE
```python
import numpy
s = numpy.array([1, 1])
f = numpy.array([False, False])
numpy.isnan(s).sum() / (f).sum()
# RuntimeWarning: invalid value encountered in long_scalars
#     numpy.isnan(s).sum() / (f).sum()
```
2. MWE
```python
import numpy
s = numpy.array([numpy.nan, numpy.nan])
numpy.nanstd(s)
# RuntimeWarning: Degrees of freedom <= 0 for slice.
#     var = nanvar(a, axis=axis, dtype=dtype, out=out, ddof=ddof,
```

### Versions of your packages

```momepy=0.5.3+13.g67d9509.dirty```

PR #389, opened 2022-09-10
handle scenarios leading to dimensions.StreetProfile() warnings

jGaboardi:
Address #388 & xref https://github.com/martinfleis/sdss22-workshop/issues/8

### DIFF
--- a/momepy/dimension.py
+++ b/momepy/dimension.py
@@ -5,7 +5,6 @@
 # definitions of dimension characters
 
 import math
-import warnings
 
 import numpy as np
 import pandas as pd
@@ -639,16 +638,19 @@ class StreetProfile:
             )
             widths.append(np.mean([left_mean, right_mean]) * 2)
 
-            with warnings.catch_warnings():
-                warnings.filterwarnings(
-                    "ignore", message="invalid value encountered in long_scalars"
-                )
-                openness.append(np.isnan(s).sum() / (f).sum())
-            with warnings.catch_warnings():
-                warnings.filterwarnings(
-                    "ignore", message="Degrees of freedom <= 0 for slice."
-                )
-                deviations.append(np.nanstd(s))
+            f_sum = (f).sum()
+            if not f_sum:
+                openness_score = np.nan
+            else:
+                openness_score = np.isnan(s).sum() / f_sum
+            openness.append(openness_score)
+
+            s_nan = np.isnan(s).all()
+            if s_nan:
+                deviation_score = np.nan
+            else:
+                deviation_score = np.nanstd(s)
+            deviations.append(deviation_score)
 
             if do_heights:
                 b = blgs[f]
--- a/momepy/dimension.py
+++ b/momepy/dimension.py
@@ -639,14 +639,15 @@ class StreetProfile:
             widths.append(np.mean([left_mean, right_mean]) * 2)
 
             f_sum = (f).sum()
+            s_nan = np.isnan(s)
+
             if not f_sum:
                 openness_score = np.nan
             else:
-                openness_score = np.isnan(s).sum() / f_sum
+                openness_score = s_nan.sum() / f_sum
             openness.append(openness_score)
 
-            s_nan = np.isnan(s).all()
-            if s_nan:
+            if s_nan.all():
                 deviation_score = np.nan
             else:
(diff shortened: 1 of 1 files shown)


martinfleis:
I'm curious if we could actually avoid the situations when these warnings are emitted rather than suppressing them. This is not really a solution of the issue (though it does resolve the sdss issue).

jGaboardi:
> I'm curious if we could actually avoid the situations when these warnings are emitted rather than suppressing them. This is not really a solution of the issue (though it does resolve the sdss issue).

Good question. I was wondering if situations 1 & 2 from #388 were valid scenarios...

martinfleis:
Yes, these may happen in reality and the result numpy gives now is okay. But it may be better to do that explicitly, not implicitly rely on numpy.

jGaboardi:
> Yes, these may happen in reality and the result numpy gives now is okay. But it may be better to do that explicitly, not implicitly rely on numpy.

Working on a solution now. Will be able to push something up soon.

jGaboardi:
Here are some conditions for catching the scenarios that lead to the warning:
```python
import numpy
s = numpy.array([1, 1])
f = numpy.array([False, False])
f_sum = (f).sum()
if not f_sum:
    openness_score = numpy.nan
else:
    openness_score = numpy.isnan(s).sum() / (f).sum()
```
```python
import numpy
s = numpy.array([numpy.nan, numpy.nan])
s_nan = numpy.isnan(s).all()
if s_nan:
    deviation_score = numpy.nan
else:
    deviation_score = numpy.nanstd(s)
```

martinfleis:
Thanks! This looks great. I assume there's no significant performance hit caused by the all nan check?

jGaboardi:
> Thanks! This looks great. I assume there's no significant performance hit caused by the all nan check?

Does not seem to be noticeable, but let me rerun again to confirm.

jGaboardi:
```python
%timeit momepy.StreetProfile(streets, buildings)
```
Running the above on cell 34 of [sdss2022 notebook](https://github.com/martinfleis/sdss22-workshop/blob/main/demo-notebook.ipynb) results in:
* `pysal/main`: 1.71 s ± 31.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
* `jGaboardi:long_scalar_warning]`: 1.92 s ± 129 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

Seems like a reasonable hit (0.2 sec.) for [a city of ≈80,000](https://en.wikipedia.org/wiki/St._Gallen).

martinfleis:
That's okay I guess. I'll try to run the code tomorrow but it seems okay now. Thanks!

momepy/dimension.py
```diff
@@ -637,8 +637,20 @@ def __init__(
...
+            else:
+                openness_score = np.isnan(s).sum() / f_sum
+            openness.append(openness_score)
+
+            s_nan = np.isnan(s).all()
+            if s_nan:
+                deviation_score = np.nan
+            else:
+                deviation_score = np.nanstd(s)
+            deviations.append(deviation_score)
```
martinfleis, line 653:
```suggestion
            f_sum = (f).sum()
            s_nan = np.isnan(s)
            if not f_sum:
                openness_score = np.nan
            else:
                openness_score = s_nan.sum() / f_sum
            openness.append(openness_score)

            if s_nan.all():
                deviation_score = np.nan
            else:
                deviation_score = np.nanstd(s)
            deviations.append(deviation_score)
```

This may be better as we call `np.isnan(s)` only once.

jGaboardi: @martinfleis good call. This was with only one `np.isnan` call, which brings the time down closer to the original. For reference:

* `pysal/main`: 1.71 s ± 31.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
* `jGaboardi:long_scalar_warning`: 1.92 s ± 129 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
  * 2 `np.isnan` calls
* `jGaboardi:long_scalar_warning`: 1.77 s ± 95 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
  * 1 `np.isnan` call

martinfleis (approved):
Thanks!

PR #393, opened 2022-09-11
DOC: create preprocessing section

martinfleis:
This creates a preprocessing section in the user guide and moves and renames existing `preprocessing` notebook. I have also added redirection from the old location of the notebook to the new one, though I am not sure it works now :). Want to see the RTD build.

@gregmaya @gsagostini please put your notebooks in this section once this gets merged.


martinfleis:
> I have also added redirection from the old location of the notebook to the new one, though I am not sure it works now :)

It does not, even though it is 1:1 copy from geopandas. Will need to debug.

jGaboardi (commented):
Two things for the `simple_preprocessing` notebook:
* Can we rerun for consecutive cell numbering?
* Correct spelling in text above cell 29:
  * "... in this situation as we are ***~~nor~~ -> not*** snapping endpoints to endpoints."

jGaboardi:
@martinfleis is this ready to merge or does it still need debugging as you mentioned above?

martinfleis:
I still need to figure out the redirect. Working on it right now.

jGaboardi (approved):

martinfleis:
Fixed, merging.

PR #395, opened 2022-09-11
handle `All NaN slice` warning

### LINKED ISSUE #390: limit numpy version evaluations in diversity.Percentiles
https://github.com/pysal/momepy/blob/67d9509c5e6a7122998e175dc47606f61f242c97/momepy/diversity.py#L901-L917

Yesterday I put the `numpy` version declaration within the for loop (line 908-911). This probably makes more sense to do it once before the food loop + declare the keyword in line 901.

xref #387

### LINKED ISSUE #394: Handle `All-NaN slice encountered` warning in diversity.Percentiles()
### Describe the problem

A `RuntimeWarning` (`All-NaN slice encountered`) is thrown from within `diversity.Percentiles()` by `numpy.nanpercentile` (see [here](https://github.com/pysal/momepy/blob/b2d88846a2a3c4162682aa7ce0ec0d4040260975/momepy/diversity.py#L913)) when [`a`](https://numpy.org/doc/stable/reference/generated/numpy.nanpercentile.html#numpy-nanpercentile) is all `numpy.nan`s. This scenario should be caught and handle prior to the `numpy.nanpercentile`, like what was done in #389. MWE below.

This warning can also potentially be thrown in [`utils.limit_range`](https://github.com/pysal/momepy/blob/67d9509c5e6a7122998e175dc47606f61f242c97/momepy/utils.py#L475-L486).

### Steps to reproduce

```python
import numpy
percentiles = [25, 50, 75]
value_arrays = [
    numpy.array([[10, numpy.nan, 0]]),                # not all NaNs
    numpy.array([[numpy.nan, numpy.nan, numpy.nan]])  # all NaNs
]
for a in value_arrays:
    a_all_nan = numpy.isnan(a).all()
    if a_all_nan:
        values_list = a
    else:
        values_list = numpy.nanpercentile(a, percentiles, method="midpoint")
    print(values_list)
# [5. 5. 5.]       # return percentiles of `a` object
# [[nan nan nan]]  # return original `a` object
```

### Versions of your packages

```momepy='0.5.3+16.gb30e8cc.dirty'```

### Your operating system

_No response_

### Additional context

_No response_

### PR DESCRIPTION
**Pull request by jGaboardi**
Resolves #394 and https://github.com/martinfleis/sdss22-workshop/issues/8

This PR catches scenarios causing `RuntimeWarning: All-NaN slice encountered` and handles appropriately. However, timing takes a pretty big hit (single column `Percentiles` instantiation for `meshedness` in cell 48 of the [sdss2022 notebook](https://github.com/martinfleis/sdss22-workshop/blob/main/demo-notebook.ipynb)):
* *no check for nans:* **2.65 s** ± 4.95 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
* *check for all nans:* **3.89 s** ± 306 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

Also, refactoring `util.limit_range()` as mentioned in #394 and addressing #390.

Closes #390

### CODE REVIEW

File: momepy/diversity.py [hunk 1]
```diff
@@ -899,19 +899,21 @@ def __init__(
...
-                        method = dict(interpolation=interpolation)
-                    results_list.append(
-                        np.nanpercentile(values_list, percentiles, **method)
-                    )
+                    all_nan = np.isnan(values_list).all()
+                    if not all_nan:
+                        values_list = np.nanpercentile(
+                            values_list, percentiles, **method
+                        )
+                    results_list.append(values_list)
```
martinfleis, line 916: What happens when `all_nan == True`? You append an array of an arbitrary length full of NaNs to `results_list`, instead of an array of length == `len(percentiles)` as `nanpercentiles` itself returns (I assume). I don't think that is the expected behaviour.

momepy/utils.py
```diff
@@ -470,20 +470,19 @@ def limit_range(vals, rng):
     array
         limited array
     """
+
     vals = np.asarray(vals)
-    if len(vals) > 2:
+
+    if len(vals) > 2 or not np.isnan(vals).all():
```
martinfleis, line 476: Didn't you mean `and` instead of `or`?

momepy/utils.py
```diff
@@ -470,20 +470,19 @@ def limit_range(vals, rng):
...
             method = dict(method="nearest")
         else:
             method = dict(interpolation="nearest")
         rng = sorted(rng)
-        if np.isnan(vals).any():
-            lower = np.nanpercentile(vals, rng[0], **method)
-            higher = np.nanpercentile(vals, rng[1], **method)
-        else:
-            lower = np.percentile(vals, rng[0], **method)
-            higher = np.percentile(vals, rng[1], **method)
```
martinfleis, line 485: Then reason being this is performance. I know you can pass anything to `nanpercentile` but there's a performance hit doing that, compared to `percentile`. At there was when I wrote this.

momepy/diversity.py
jGaboardi: Ah! Good catch! How about this:

```python
import numpy
percentiles = [25, 75]
value_arrays = [
    numpy.array([[10, 5, 0]]),                        # all numeric
    numpy.array([[10, numpy.nan, 0]]),                # not all NaNs
    numpy.array([[numpy.nan, numpy.nan, numpy.nan]])  # all NaNs
]
nan_filler = numpy.empty(len(percentiles))
nan_filler[:] = numpy.nan
for a in value_arrays:
    a_all_nan = numpy.isnan(a).all()
    if a_all_nan:
        values_list = nan_filler
    else:
        values_list = numpy.nanpercentile(a, percentiles, method="midpoint")
    print(values_list)
# [2.5 7.5]       # return percentiles of `a` object
# [5. 5.]         # return original `a` object
# [nan nan]       # return array length len(percentiles) filled with NaNs
```

momepy/utils.py
jGaboardi: Ah, OK. I'll do more timing tests.

momepy/utils.py
jGaboardi: Yes, I think my brain is a bit foggy.

As a side note, why not `len(vals) > 1`?

momepy/utils.py
jGaboardi: Yep, using `nanpercentile` regardless of NaNs in the collection makes a performance difference. Good call.

```python
import momepy as mm
import numpy as np

%timeit mm.limit_range(np.array([0, 1, 2]), rng=(25, 75))
# 99.9 µs ± 5.32 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

%timeit mm.limit_range(np.array([np.nan, 1, 2]), rng=(25, 75))
# 115 µs ± 3.71 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

%timeit mm.limit_range(np.array([1, 2]), rng=(25, 75))
# 598 ns ± 12.9 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

%timeit mm.limit_range(np.array([np.nan, np.nan, np.nan]), rng=(25, 75))
# 2.69 µs ± 89.1 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

jGaboardi:
Here's a timing update for `diversity.Percentiles()`:
* 2.65 s ± 4.95 ms per loop (mean ± std. dev. of 7 runs, 1 loop each) ---- no check for nans
* 3.89 s ± 306 ms per loop (mean ± std. dev. of 7 runs, 1 loop each) ---- check for all nans (wrong shape)
* 3.31 s ± 25.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each) ---- nan filler (correct shape)

momepy/utils.py
```diff
@@ -470,8 +470,10 @@ def limit_range(vals, rng):
     array
         limited array
     """
+
     vals = np.asarray(vals)
-    if len(vals) > 2:
+
+    if (len(vals) > 2) and (not np.isnan(vals).all()):
```
martinfleis, line 476: Cany you assing `np.isnan(vals)` to a variable and check against it? Because we do isnan here and then on line 482 again, so we can do it once and then do `var.all()` and `var.any()`.

martinfleis:
> Here's a timing update

does this suggest that the best solution would be to keep the implementation we have and just suppress the warning?

momepy/utils.py
jGaboardi: Tried this and it appears no significant timing change.

jGaboardi:
> > Here's a timing update
> 
> does this suggest that the best solution would be to keep the implementation we have and just suppress the warning?

Seems like yes. Let me test the warning timing again to make sure.

momepy/utils.py
jGaboardi: Declaring a variable for `var = np.isnan(vals)` does not seem to make a significant difference in timing, but it also doesn't hurt. Probably should go with variable declaration?

jGaboardi:
> > > Here's a timing update
> > 
> > 
> > does this suggest that the best solution would be to keep the implementation we have and just suppress the warning?
> 
> Seems like yes. Let me test the warning timing again to make sure.

So the answer is 'yes,' simply catching the `All-NaN slice encountered` warning is the most efficient solution currently (nearly 1 second faster for one column in the sdss2022 notebook).

* 2.51 s ± 14.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

Shall I push that up for the code chunk in `Percentiles()` and update `limit_range()` with the actual `nump.isnan()` checker?

martinfleis:
Okay. Sounds good.

jGaboardi:
> Okay. Sounds good.

Still feel like I might be missing something. Going try out a couple more things.

martinfleis:
I think you haven't pushed.

jGaboardi:
Weird. [CI ran for the push](https://github.com/pysal/momepy/actions/runs/3038949553/jobs/4893299442#step:5:160) and the [files changes are current](https://github.com/pysal/momepy/pull/395/files), but no commit/run report here in the actual PR.

martinfleis (approved):
Thanks! Merge when green.

martinfleis:
Just a tip - use [closing keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) in the first comment instead of "addressing" or similar. Those will automatically close the original issue(s) when merged.

PR #400, opened 2022-09-12
standardize pytest.approx() in testing files

### LINKED ISSUE #399: both pytest.approx() and approx() used in testing
### Describe the problem

Both `pytest.approx()` and `approx()`are used in testing, sometimes within the same testing files as in `test_dimension.py`.

https://github.com/pysal/momepy/blob/820d80e5a9a58dbe46d4e56e10ec7017835d672e/tests/test_dimension.py#L4-L5

We should choose 1 and stick with it. I propose the following pattern:

```python
import pytest
assert thing1 == pytest.approx(thing2, rel=1e-3)
```

### Steps to reproduce

n/a

### Versions of your packages

```momepy=0.5.3+18.g2a1b209```

### Your operating system

_No response_

### Additional context

_No response_

jGaboardi:
Closes #399.

### DIFF
--- a/tests/test_dimension.py
+++ b/tests/test_dimension.py
@@ -2,7 +2,6 @@ import geopandas as gpd
 import numpy as np
 import pandas as pd
 import pytest
-from pytest import approx
 from shapely.geometry import LineString, Point, Polygon
 
 import momepy as mm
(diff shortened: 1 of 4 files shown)


martinfleis (approved):
Thanks

PR #406, opened 2022-09-13
API: deprecation decorators for transition to functional API

martinfleis:
This is a proof-of-concept of the switch from the class-based API we have now to the functional API discussed in #340. It illustrates before-after as well as the transition period. 

A couple of things to discuss:

1. current classes allow users to retrieve all objects used in it, from the original gdf to weights or anything that has been created ad-hoc. We may want to retain such a behaviour but it leads to some code duplication in the meantime. Thoughts?
2. not sure if the current behaviour of the warning is not too noisy. It will show up every time you use class based API.
3. I would like to use this change to cleanup the arguments a bit as well, to get the naming consistency with the rest of PySAL (e.g. `spatial_weights -> w`) and potentially even remove some stuff from momepy and move it elsewhere, like all our fancy ways of building lags (e.g. [AverageCharacter](http://docs.momepy.org/en/stable/generated/momepy.AverageCharacter.html#momepy.AverageCharacter)).
4. In most cases, W now depends on a `unique_id` column, that has been used to index the weights. I would like to get rid of that and depend on the actual index of gdf instead, or at least have this controlled by a keyword. In most cases, passing `unique_id` is completely unnecessary. That all depends on the correct behaviour of libpysal in the first place, though (xref https://github.com/pysal/libpysal/pull/477). As a result, a lot would break but the final API in 1.0 will be much cleaner and easier to use.

@jGaboardi thoughts? It is a lot of breakage but if it happens all at once with a proper migration guide, it may not hurt that much. And since we're following semver here, it still acceptable.


jGaboardi:
tl;dr – I like it.

1. IMHO temporary code duplication is acceptable; just as long as we prune it down as soon as possible.
2. I think it is as noisy as it has to be to get people to switch to the functional API sooner than later. Maybe we add a manual setting for users that *really* don't want to make the change? Possibly keyword to skip the [warning](https://github.com/pysal/momepy/pull/406/files#diff-e0b677baf83b1a1339d5a964645035870cb0d00d228781660899e2effe5f85fdR33) or maybe `global` that can be switched on.
3. Very much on board for this.
4. Also on board for this.

martinfleis:
I am going to revise this and, on top of the points listed above, build the new implementation on top of `Graph`.

martinfleis:
I have updated this to reflect the current state of things. I'd like to merge it so we can use it as some classes are to be removed and adding a decorator would be nice to do now while we remember that.

momepy/utils.py
```diff
@@ -18,6 +18,57 @@
...
+    def decorator(func1):
+        import functools
+
+        @functools.wraps(func1)
+        def new_func1(*args, **kwargs):
+            warnings.warn(
+                f"Class based API like `momepy.{func1.__name__}` is deprecated. "
+                f"Replace it with `momepy.{new_way}` to use functional API instead "
+                "or pin momepy version <1.0. Class-based API will be removed in 1.0. "
+                # "See details at https://docs.momepy.org/en/stable/migration.html",
```
jGaboardi, line 35: Should this line be uncommented? (Also L60 below)?

martinfleis: No, as we don't have the migration guide written yet :)

jGaboardi (approved):

PR #418, opened 2022-09-19
GHA: update PyPI release action

martinfleis:
The GHA is complaining over at geopandas/xyzservices where we have the same config:

> [Build and publish xyzservices to PyPI: # >> PyPA publish to PyPI GHA: UNSUPPORTED GITHUB ACTION VERSION <<#L1](https://github.com/geopandas/xyzservices/commit/e6315d8026569fa7626a265ae9d89642d4a75aae#annotation_4652189477)
You are using "pypa/gh-action-pypi-publish@master". The "master" branch of this project has been sunset and will not receive any updates, not even security bug fixes. Please, make sure to use a supported version. If you want to pin to v1 major version, use "pypa/gh-action-pypi-publish@release/v1". If you feel adventurous, you may opt to use use "pypa/gh-action-pypi-publish@unstable/v1" instead. A more general recommendation is to pin to exact tags or commit shas.

this should fix that

### DIFF
--- a/.github/workflows/release_to_pypi.yml
+++ b/.github/workflows/release_to_pypi.yml
@@ -25,7 +25,7 @@ jobs:
           python setup.py sdist bdist_wheel
 
       - name: Publish distribution to PyPI
-        uses: pypa/gh-action-pypi-publish@master
+        uses: pypa/gh-action-pypi-publish@release/v1
         with:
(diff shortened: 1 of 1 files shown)


martinfleis:
We may need to deploy the James Bot to update this across PySAL :).

Issue #423, opened 2022-09-20
pandas 1.5.0 compatibility issues


martinfleis:
pandas 1.5 released yesterday broke our CI and is emitting a ton of new deprecation warnings.

https://github.com/pysal/momepy/actions/runs/3091088424/jobs/5000737392#step:6:187

jGaboardi:
Not 100% sure why [this is complaining](https://github.com/pysal/momepy/actions/runs/3091088424/jobs/5000737392#step:6:163) so much. The message indicates the problem is with using `.iloc`, but we are using [`.loc` there](https://github.com/pysal/momepy/blob/main/momepy/elements.py#L281). I had thought using `.loc` was always safe for setting values in place?

```
elements.py:281: FutureWarning: In a future version, `df.iloc[:, i] = newvals` will attempt to set the values inplace instead of always setting a new array. To retain the old behavior, use either `df[df.columns[i]] = newvals` or, if columns are non-unique, `df.isetitem(i, newvals)`
```

martinfleis:
The warning is a bit confusing but the change seem to apply to both `loc` and `iloc` - https://pandas.pydata.org/docs/whatsnew/v1.5.0.html#inplace-operation-when-setting-values-with-loc-and-iloc

martinfleis:
The Percentiles broken in the SDSS 2022 tutorial are caused by pandas 1.5.0. With 1.4.4 as the only difference between the envs (numpy same in both), the issue is not present.

jGaboardi:
Probably in [`numpy.nanpercentile`](https://github.com/pysal/momepy/blob/c574bcfd8445dc8b5e54912cd8fe28ea10ff2cf5/momepy/diversity.py#L904-L923), correct? Or is it specifically in `pandas`?

martinfleis:
Update - there is a mismatch in assigned `unique_id` in buildings and tessellation. The building and its tessellation cell should have the same id. They do not right after tessellation is created:

![Screenshot 2022-09-21 at 21 56 24](https://user-images.githubusercontent.com/36797143/191598543-e712de99-8203-4335-96fc-e1a22fd200b6.png)

So the issue seems to be within `Tessellation`. Digging deeper.

martinfleis:
One note aside - we have a bug in Percentiles as well. This should be the same as line 898 https://github.com/pysal/momepy/blob/c574bcfd8445dc8b5e54912cd8fe28ea10ff2cf5/momepy/diversity.py#L925

[...]

PR #427, opened 2022-09-21
resolves bug in Percentiles append.

### LINKED ISSUE #426: bug in Percentiles append
One note aside - we have a bug in Percentiles as well. This should be the same as line 898 https://github.com/pysal/momepy/blob/c574bcfd8445dc8b5e54912cd8fe28ea10ff2cf5/momepy/diversity.py#L925

_Originally posted by @martinfleis in https://github.com/pysal/momepy/issues/423#issuecomment-1254168703_

jGaboardi:
Resolves #426.

### DIFF
--- a/momepy/diversity.py
+++ b/momepy/diversity.py
@@ -922,7 +922,7 @@ class Percentiles:
                             np.nanpercentile(values_list, percentiles, **method)
                         )
                 else:
-                    results_list.append(np.nan)
+                    results_list.append(np.array([np.nan] * len(percentiles)))
 
             self.frame = pd.DataFrame(
                 results_list, columns=percentiles, index=gdf.index
--- a/tests/test_diversity.py
+++ b/tests/test_diversity.py
@@ -1,6 +1,8 @@
 import geopandas as gpd
 import numpy as np
+import pandas as pd
 import pytest
+from shapely.geometry import Polygon
 
 import momepy as mm
 from momepy import sw_high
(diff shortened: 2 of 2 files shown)


martinfleis:
We should probably test this somehow.

jGaboardi:
> We should probably test this somehow.

I've added two tests (one unweighted and one weighted) to check for correct behavior.

jGaboardi:
xref #423 Interesting that now `py38` tests are passing here where `py39`, etc. fail. The only major difference that I found is [`numpy=1.21.6`](https://github.com/pysal/momepy/actions/runs/3102334399/jobs/5024541151#step:5:161) vs. [`numpy=1.23.3`](https://github.com/pysal/momepy/actions/runs/3102334399/jobs/5024541264#step:5:162). Seems at least possibly related to the weird [`numpy.argsort` behavior](https://github.com/geopandas/geopandas/issues/2558#issuecomment-1254325817) that @jorisvandenbossche found deep in `pandas`?

jGaboardi:
xref https://github.com/pandas-dev/pandas/pull/48711

jGaboardi:
@martinfleis Anything more you think should be done for this?

martinfleis (approved):
I was hoping to see the green ci but given pandas release may not come soon, let's merge.

I am thinking we should do a patch release that changes our requirements to avoid ppl running into that bug...

Issue #437, opened 2022-10-08
patch release pinning pandas=1.4.4?


jGaboardi:
xref https://github.com/pysal/momepy/pull/427#pullrequestreview-1127124471

> I am thinking we should do a patch release that changes our requirements to avoid ppl running into that bug...

@martinfleis Should we go ahead and pin [`pandas=1.4.4`](https://github.com/pandas-dev/pandas/releases/tag/v1.4.4) for a patch release until the actual fix in `pandas` is available?

jGaboardi:
This can probably closed now that [`pandas==1.5.1`](https://pandas.pydata.org/pandas-docs/version/1.5.1/whatsnew/v1.5.1.html#fixed-regressions) is out, right?

martinfleis:
yes. and it was already resolve https://github.com/pysal/momepy/pull/438

jGaboardi:
LOL. Whoops.

PR #438, opened 2022-10-08
Pin pandas versions prior to patch release

jGaboardi:
xref #437, #423

* pin `pandas==1.4.4` for most CI and set up environments
* pin `pandas>=0.230,<=1.4.4` for minimal environment


martinfleis:
Can we just say any pandas except 1.5.0 instead of pinning to a specific one?

martinfleis:
Thanks!

Issue #294, opened 2021-08-10
Documentation doesn't build a pdf


martinfleis:
RTD refuses to build a PDF because of the huge amount of numpy deprecation warnings in there (https://github.com/geopandas/geopandas/issues/2013). I have disabled the option for now but we should re-enable it once it gets fixed in geopandas 0.10.0.

jGaboardi:
@martinfleis Is this fixed now?

martinfleis:
Don't know, we need to revert https://github.com/pysal/momepy/commit/3dd803b566d0a048c140b3c2d1be44f4fa8fb2f8 to check.

jGaboardi:
> we should re-enable it once it gets fixed in geopandas 0.10.0.

Since `geopandas==0.10.0` has been out for a while now, I will see how this behaves now.

martinfleis:
Still not working. RTD is not event trying.

jGaboardi:
The [build is now failing](https://github.com/pysal/momepy/pull/455#issuecomment-1375703120), but at least it's running. Since the original problem was with the amount of `numpy` warning. I am 99% sure the current failure is due to the amount of `pandas` warnings for clip, .iloc, etc.

jGaboardi:
That's a "Martin" decision in the end. I can see value in having it for those who would like access to rich docs in a secure data center or the like (as I have desperately desired in the past), but this is certainly an edge case. I will defer to your judgment on it. Maybe we even punt for now and try again once all the `pandas` warnings get worked out?

martinfleis:
> Maybe we even punt for now and try again once all the pandas warnings get worked out?

+1

not really keen spending a ton of energy on this

PR #455, opened 2023-01-08
retry building PDF from `readthedocs.yml`

jGaboardi:
This PR reverts 3dd803b. We may also want to try update the Python version with `readthedocs.yml`.

xref #294

### DIFF
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -1,5 +1,5 @@
 version: 2
-formats: []
+formats: [pdf]
 
 sphinx:
   configuration: docs/conf.py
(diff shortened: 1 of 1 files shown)


jGaboardi:
Failures are not due to this PR; see #456 & #457.

martinfleis:
How do we know it is okay now? This https://readthedocs.org/projects/momepy/builds/19107913/ doesn't seem to build a PDF. Maybe that is enabled only on main?

jGaboardi:
I was about to ask you that question... 😆

martinfleis:
had to enable it in RTD

jGaboardi:
Although it was triggered, the build [failed](https://readthedocs.org/projects/momepy/builds/19111373/).

PR #467, opened 2023-02-23
MAINT: switch to pyproject and ruff

martinfleis:
Updating infra as per https://github.com/pysal/submodule_template/pull/41 (roughly).


.github/workflows/tests.yaml
```diff
@@ -4,67 +4,63 @@ on:
...
     branches: [main]
   pull_request:
-    branches: [main]
+    branches:
+      - "*"
   schedule:
     - cron: "0 0 * * 1,4"
 
 jobs:
-  Linting:
```
jGaboardi, line 12: So we will no longer be doing a formatting/linting check in CI?

martinfleis: I have enabled pre-commit.ci instead. See the checks below.

jGaboardi (approved):

PR #468, opened 2023-02-25
remove redundant `requirements.txt`

jGaboardi:
This PR follows up with #467
* removes `--cov-config .coveragerc` from `tests.yml` (there's no longer a `.coveragerc` file`)
* removes the redundant `requirements.txt`, which now reside within [`pyproject.toml`](https://github.com/pysal/momepy/blob/b4a415677fe3b2c0f9ce5f9700833cb88fae583f/pyproject.toml#L36-L44).

https://github.com/pysal/momepy/blob/b4a415677fe3b2c0f9ce5f9700833cb88fae583f/pyproject.toml#L36-L44

*

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -46,7 +46,7 @@ jobs:
 
       - name: Test momepy
         run: |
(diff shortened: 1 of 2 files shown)


martinfleis (approved):
Nice catch!

Issue #472, opened 2023-03-12
TypeError when using momepy.remove_false_nodes with shapely.geometry.linestring.LineString


anisotropi4:
### Describe the problem

The `momepy.remove_false_nodes` function call errors with a `TypeError` when called with a `GeoSeries` `LineString` created with `GeoPandas 0.12.2` when using `Shapely 2.0`.

```python
momepy.remove_false_nodes(line)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/dev/momepy/venv/lib/python3.10/site-packages/momepy/preprocessing.py", line 224, in remove_false_nodes
    coords = pygeos.get_coordinates(geom)
  File "/home/dev/momepy/TIPLOC/venv/lib/python3.10/site-packages/pygeos/coordinates.py", line 134, in get_coordinates
    return lib.get_coordinates(
TypeError: One of the arguments is of incorrect type. Please provide only Geometry objects.
```

### Steps to reproduce

Executing the following raises a `TypeError`:
```python   
import os
os.environ['USE_PYGEOS'] = '0'
import geopandas
import momepy
from shapely import LineString
line = geopandas.GeoSeries(LineString([[0, 1], [2, 3]]))
momepy.remove_false_nodes(line)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/dev/momepy/venv/lib/python3.10/site-packages/momepy/preprocessing.py", line 224, in remove_false_nodes
    coords = pygeos.get_coordinates(geom)
  File "/home/dev/momepy/TIPLOC/venv/lib/python3.10/site-packages/pygeos/coordinates.py", line 134, in get_coordinates
    return lib.get_coordinates(
TypeError: One of the arguments is of incorrect type. Please provide only Geometry objects.
```

Whereas the following raising no error:
```python   
import os
os.environ['USE_PYGEOS'] = '1'
import geopandas
import momepy
from shapely import LineString
line = geopandas.GeoSeries(LineString([[0, 1], [2, 3]]))
momepy.remove_false_nodes(line)
```

### Versions of your packages

```
momepy=0.5.4
geopandas=0.12.2
libpysal=4.7.0
networkx=3.0
```

### Your operating system

Linux 5.15.0-60-generic #66-Ubuntu x86_64 GNU/Linux

### Additional context

[...]

Issue #443, opened 2022-10-24
Refactor code for shapely 2.0


martinfleis:
We have some parts of the code depending on pygeos. Now that shapely 2.0b1 is out and GeoPandas is on its way to remove pygoes backend (and shapely 1.8 backend as well), we need to make sure momepy is ready.

We don't have to do anything right now, but when the time comes, we need to be able to work with shapely instead of pygeos. 

I don't really like supporting both at the same time, so we should probably have a discussion how to do such a transition. GeoPandas will have a feature parity with shapely 2.0 before it will remove pygeos, so my suggestion is to wait for that and use geopandas methods instead of shapely/pygeos functions then. That allows us to pin min geopandas version and forget about anything else.

PR #473, opened 2023-03-13
TypeError when using momepy.remove_false_nodes with shapely.geometry.linestring.LineString

anisotropi4:
Use `USE_PYGEOS` environment variable to import `shapely` or `pygeos` functions as `geos` functions.

Apologies as this should be under "**TypeError when using momepy.remove_false_nodes with shapely.geometry.linestring.LineString #472**"

The patch tests the value of the `USE_PYGEOS` environment variable. If set to '0' it imports `shapely` functions as the `geos` package otherwise it imports `pygeos` functions as the `geos` package.

The `tree.query` and `tree.query_bulk` functions are also wrapped as the `shapely` `tree.query` function is equivalent to the `pygeos` `tree.bulk_query` function when passed an array.

It arguably maintains backward compatibility and passes my local tests and the `test/test_preprocess.py` `pytest` module

### DIFF
--- a/momepy/preprocessing.py
+++ b/momepy/preprocessing.py
@@ -6,13 +6,6 @@ import operator
 import warnings
 import os
 
-PYGEOS = True
-if os.environ.get("USE_PYGEOS", "1") == "0":
-    PYGEOS = False
-    import shapely as geos
-else:
-    import pygeos as geos
-
 import geopandas as gpd
 import libpysal
 import numpy as np
@@ -37,6 +30,15 @@ __all__ = [
 
 GPD_10 = Version(gpd.__version__) >= Version("0.10")
 GPD_09 = Version(gpd.__version__) >= Version("0.9")
+
+if os.environ.get("USE_PYGEOS", "1") == "1":
+    raise DeprecationWarning(
+        """The environment variable USE_PYGEOS is either 
+    unset or set to 1. Due to breaking changing this module has been updated to
+    use Shapely 2.0 which may make some functions fail. To use and test Shapely
+    2.0, you have set the environment variable USE_PYGEOS=0.
+    """
+    )
 
 
 def preprocess(
@@ -228,8 +230,8 @@ def remove_false_nodes(gdf):
         df = gpd.GeoSeries(gdf)
 
     # extract array of coordinates and number per geometry
-    coords = geos.get_coordinates(geom)
-    indices = geos.get_num_coordinates(geom)
+    coords = shapely.get_coordinates(geom)
+    indices = shapely.get_num_coordinates(geom)
 
     # generate a list of start and end coordinates and create point geometries
     edges = [0]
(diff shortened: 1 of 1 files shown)


anisotropi4:
This appear to break the `Read the Docs` build in a way that I don't understand. Help to resolve this would be appreciated.

martinfleis:
Hey, thanks for the efforts but we can't resolve it in this way. This would override env var set by a user and usage of shapely backend would break other functions we have here. We need to tackle whole #443 at one go following this approach https://github.com/geopandas/geopandas/issues/2691#issuecomment-1375890344.

anisotropi4:
> Hey, thanks for the efforts but we can't resolve it in this way. This would override env var set by a user and usage of shapely backend would break other functions we have here. We need to tackle whole #443 at one go following this approach [geopandas/geopandas#2691 (comment)](https://github.com/geopandas/geopandas/issues/2691#issuecomment-1375890344).

Hey, no worries. `momepy` and the other stuff here is great and as a dilettante I find it really useful for my weekend stuff.

Given this, I'm will question this comment as the point about this patch is that it *is intended to use the` env var` set by a user not override the `env var*. I get it does some import stuff based on the value instead so maintain some kind of backwards compatibility.

It is an ugly kludge and that this should be done properly but for me it is a blocking issue. Code that had worked with `geopandas` and `momepy` randomly broke in a new and mysterious way on the weekend. The `TypeError` in `momepy` wasn't the most obvious as to why this was happening. Hence the patch.

Equally, if backward compatibility isn't an issue, then for the `preprocess.py` module at least, replace all `pygeos` references with `shapely` and update `tree.bulk_query` with `tree.query` and have done with it.

martinfleis:
> Equally, if backward compatibility isn't an issue, then for the preprocess.py module at least, replace all pygeos references with shapely and update tree.bulk_query with tree.query and have done with it.

The approach I linked above is only partially backwards compatible. Being compatible with shapely 1.8 is a lot of hassle and I'd rather not maintain that kind of code here and push towards a faster adoption of shapely 2.0.

In these functions you are interested in, you will need to replace pygeos functions with shapely ones and use `geom = df.geometry.values` instead of `geom = df.geometry.values.data` to ensure that either backend (pygeos or shapely) will work. 

The `import shapely as geos` is also not needed, you should be able to use `shapely` functions on a GeometryArray no matter if pygeos is used under the hood or not.

momepy/preprocessing.py
```diff
@@ -4,12 +4,19 @@
 import math
 import operator
 import warnings
+import os
+
+PYGEOS = True
+if os.environ.get("USE_PYGEOS", "1") == "0":
```
martinfleis, line 10: You can't expect True if the variable is not set.

anisotropi4:
> > Equally, if backward compatibility isn't an issue, then for the preprocess.py module at least, replace all pygeos references with shapely and update tree.bulk_query with tree.query and have done with it.
> 
> The approach I linked above is only partially backwards compatible. Being compatible with shapely 1.8 is a lot of hassle and I'd rather not maintain that kind of code here and push towards a faster adoption of shapely 2.0.
> 
> In these functions you are interested in, you will need to replace pygeos functions with shapely ones and use `geom = df.geometry.values` instead of `geom = df.geometry.values.data` to ensure that either backend (pygeos or shapely) will work.
> 
> The `import shapely as geos` is also not needed, you should be able to use `shapely` functions on a GeometryArray no matter if pygeos is used under the hood or not.

Thanks for the steer. The `import shapely as geos` was an ugly hack for backwards compatibility.

momepy/preprocessing.py
anisotropi4: I had forgotten I had this set as a global `env var` in my `.profile`. It is academic anyway and will close this now.

anisotropi4:
I will close this now following as @martinfleis has kindly suggested fixes that would address my current issues and as a more general approach is being  the https://github.com/geopandas/geopandas/issues/2691#issuecomment-1375890344 I will leave this.

The final patch has the suggested `shapely` updates applied and passes `pytest` validation with the `env variable` `USE_PYGEOS=0`. 

It raises a `DeprecationWarning` if `USE_PYGEOS` is either unset or set to 1. Which will break things in the CI toolchain.

Thanks for the help and suggestions and I will leave it here.

PR #479, opened 2023-04-08
REF: update for shapely 2.0

### LINKED ISSUE #443: Refactor code for shapely 2.0

martinfleis:
Replacing pygeos with shapely 2.0 and requiring it to run momepy. Shall we compatible with both pygeos and shapely backed in geopandas.

Closes #443

### DIFF
--- a/ci/envs/310-latest.yaml
+++ b/ci/envs/310-latest.yaml
@@ -10,7 +10,7 @@ dependencies:
   - networkx
   - packaging
   - pandas!=1.5.0
-  - shapely>=1.8.0
+  - shapely>=2
   - tqdm
   # testing
   - codecov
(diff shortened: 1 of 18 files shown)


jGaboardi (approved):
LGTM. Great work on it. Looking forward to lots of shapely 2.0 fun!

PR #485, opened 2023-05-01
RLS: add 0.6.0 changelog

martinfleis:
@jGaboardi I am including here only stuff that may interest end-user but the commit history contains more changes (refactoring, compatibility, CI, docs) that are not listed. Do we want to include those?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,22 @@
 Changelog
 =========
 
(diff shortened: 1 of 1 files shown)


jGaboardi:
Probably not since that's more for our records. And we can look stuff up through commit history if we need to revisit any infrastructure modifications.

jGaboardi (approved):

PR #493, opened 2023-06-30
REF: Simplify list and dict manipulation

eumiro:
Only straightforward refactoring, no tests/functionality affected.


jGaboardi:
@eumiro Thank you for your contribution!

@martinfleis This way is much cleaner. Do we think there is a reason keeping the original, more verbose implementation?

martinfleis:
No reason. I was young and didn't know what I was doing :D.

martinfleis (approved):
Thanks!

PR #494, opened 2023-06-30
Modernize code

eumiro:
Get rid of a few Python2 relics (e.g. explicit floats where integers are more appropriate) and make the code again a bit more compact and (hopefully) readable.

### DIFF
--- a/momepy/coins.py
+++ b/momepy/coins.py
@@ -188,16 +188,16 @@ class COINS:
             # both the keys. The key is already present in the dictionary so
             # it does not calculate a second time.
             for link1 in self.unique[edge][2]:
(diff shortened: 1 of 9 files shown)


jGaboardi (approved):
Conditional approval based on a green CI suite.

Thanks for another contribution, @eumiro!

martinfleis (approved):
Thanks a ton! Do you have some tool pointing these out or are you just going through the code by hand?

eumiro:
There are tools like `pyupgrade`, but also a few words I grep for and look around. Python2 relics are easily recognizable around `int(`, `float(`, `.0`.

PR #495, opened 2023-07-03
TST: fix a flaky test in Count

martinfleis:
We have a CI in dev env. The result is correct but the order of rows in blocks is coming different from some geopandas operation. 

We don't really care about that so I am just replacing the test with something a bit more robust.

I also noticed that the whole `Blocks` algo would deserve a bit of refactoring but that is a problem for another time.

### DIFF
--- a/momepy/tests/test_intensity.py
+++ b/momepy/tests/test_intensity.py
@@ -3,6 +3,7 @@ import numpy as np
 import pytest
 from libpysal.weights import Queen
 from shapely.geometry import Point
+from pandas.testing import assert_series_equal
 
 import momepy as mm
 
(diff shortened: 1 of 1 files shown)


momepy/tests/test_intensity.py
```diff
@@ -86,9 +87,14 @@ def test_Count(self):
         weis = mm.Count(
             self.df_streets, self.df_buildings, "nID", "nID", weighted=True
         ).series
-        check_eib = [13, 14, 8, 26, 24, 17, 23, 19]
+        check_eib = (
+            self.df_buildings.drop(columns="bID")
+            .sjoin(self.blocks)["bID"]
```
martinfleis, line 92: Count is attribute-based, so verifying against a spatial equivalent.

jGaboardi (approved):

PR #507, opened 2023-08-19
adds multipolygon support for Squareness and CentroidCorners

songololo:
Is there appetite to support MultiPolygons for Squareness and CentroidCorners?

### DIFF
--- a/momepy/shape.py
+++ b/momepy/shape.py
@@ -903,7 +903,7 @@ class Squareness:
     """
     Calculates the squareness of each object in a given GeoDataFrame. Uses only
     external shape (``shapely.geometry.exterior``), courtyards are not included.
-     Returns ``np.nan`` for MultiPolygons.
+    Returns ``np.nan`` for MultiPolygons if containing multiple geoms.
 
     .. math::
         \\mu=\\frac{\\sum_{i=1}^{N} d_{i}}{N}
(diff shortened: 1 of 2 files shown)


martinfleis:
thanks for this! 

I understand how useful it could be in some cases but am not sure if it doesn't bring a bit opaque inconsistency when it does work for some MultiPolygons but not for the other.

@jGaboardi what are your thoughts?

jGaboardi:
> I understand how useful it could be in some cases but am not sure if it doesn't bring a bit opaque inconsistency when it does work for some MultiPolygons but not for the other.

It seems to me that the opaque inconsistency can be cleared up with several more words in the docstring (and *maybe* a thrown warning silenced by default at the absolute most). My reasoning here is that the MultiPolygons that are being operated on are not true MultiPolygons, they are single Polygons mislabeled (perhaps not the best terminology) as being MultiPolygons. Maybe something like:

```
Returns ``np.nan`` for true MultiPolygons (containing multiple geometries).
MultiPolygons with a singular geometry are treated as Polygons.
```



And tests failing due to #508, not this PR, correct?

martinfleis:
Okay, let's give it a shot following @jGaboardi's suggestion.

> And tests failing due to https://github.com/pysal/momepy/issues/508, not this PR, correct?

Yes.

momepy/shape.py
```diff
@@ -903,7 +903,7 @@ class Squareness:
     """
     Calculates the squareness of each object in a given GeoDataFrame. Uses only
     external shape (``shapely.geometry.exterior``), courtyards are not included.
-     Returns ``np.nan`` for MultiPolygons.
+    Returns ``np.nan`` for MultiPolygons if containing multiple geoms.
```
jGaboardi, line 906:
```suggestion
    Returns ``np.nan`` for true MultiPolygons (containing multiple geometries).
MultiPolygons with a singular geometry are treated as Polygons.
```

momepy/shape.py
```diff
@@ -1117,7 +1112,7 @@ def __init__(self, gdf):
 class CentroidCorners:
     """
     Calculates the mean distance centroid - corners and standard deviation.
-    Returns ``np.nan`` for MultiPolygons.
+    Returns ``np.nan`` for MultiPolygons if containing multiple geoms.
```
jGaboardi, line 1115:
```suggestion
    Returns ``np.nan`` for true MultiPolygons (containing multiple geometries).
MultiPolygons with a singular geometry are treated as Polygons.
```

momepy/shape.py
```diff
@@ -948,41 +948,36 @@ def _angle(a, b, c):
 
             return angle
 
+        def _calc(geom):
+            angles = []
+            points = list(geom.exterior.coords)  # get points of a shape
+            if len(points) < 3:
```
jGaboardi, line 954: It's small optimization, but since we are using `len(points)` 3 times here, let's declare `npoints = len(points)` once and use that here and in lines 956 & 958.

momepy/shape.py
```diff
@@ -1173,55 +1168,47 @@ def true_angle(a, b, c):
                 return True
             return False
 
+        def _calc(geom):
+            distances = []  # set empty list of distances
+            centroid = geom.centroid  # define centroid
+            points = list(geom.exterior.coords)  # get points of a shape
+            stop = len(points) - 1  # define where to stop
```
jGaboardi, line 1175: Since we're only using `len(points)` twice here, I'm on the fence for declaring that value. I'd say one argument for it is consistency with the `_calc()` implementation in `Squareness` (if we go with my suggestion). You can be the tie-breaker here, @martinfleis.

martinfleis: I don't mind either way, it is cheap.

PR #509, opened 2023-08-21
TST: adapt tests to GEOS 3.12

### LINKED ISSUE #508: TST: update for GEOS 3.12
GEOS 3.12 fixed their algo for minimum bounding rectangle causing a failure in our CI.

martinfleis:
Closes #508

### DIFF
--- a/momepy/tests/test_distribution.py
+++ b/momepy/tests/test_distribution.py
@@ -125,7 +125,7 @@ class TestDistribution:
         test_value = (
             22.744936872392813
             if Version(shapely.geos_version_string) < Version("3.12")
-            else 18.299481296
+            else 18.29842494
         )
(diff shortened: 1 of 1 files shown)


jGaboardi:
Looks like we'll want to [decrease the precision a bit to get the test passing](https://github.com/pysal/momepy/actions/runs/5929795726/job/16078211800?pr=509#step:5:249).

martinfleis:
I wasn't able to get new geos on an arm Mac so I used the original value we had prior shapely 2.0. I just pushed the fix, hopefully (can't test locally).

jGaboardi:
> I wasn't able to get new geos on an arm Mac so I used the original value we had prior shapely 2.0. I just pushed the fix, hopefully (can't test locally).

Looks like that's doing the trick.

jGaboardi (approved):

PR #516, opened 2023-11-07
GHA: update actions

martinfleis:
@jGaboardi how do I make dependabot to do this?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every week
+      interval: "weekly"
--- a/.github/workflows/release_to_pypi.yml
+++ b/.github/workflows/release_to_pypi.yml
@@ -11,10 +11,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@master
+      - uses: actions/checkout@v4
 
       - name: Set up Python
-        uses: actions/setup-python@v3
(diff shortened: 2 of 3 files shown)


jGaboardi:
In `spaghetti`, `spopt`, etc. we have a [`.github/dependabot.yml`](https://github.com/pysal/spaghetti/blob/main/.github/dependabot.yml). However, the link for configuration help in the top comments[^1] does not appear to work anymore (or at least right now). And come to think of it, I haven't seen any `dependabot` requests recently. 

[^1]: https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates

jGaboardi:
This link seems to be what we are looking for: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates

jGaboardi:
As far I understand, the versioning will be checked (e.g. `actions/setup-python`), but something like `mamba-org/provision-with-micromamba` --> `mamba-org/setup-micromamba` will not be picked up.

jGaboardi:
And we also have to [turn it on](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates#enabling-dependabot-version-updates).

martinfleis:
> As far I understand, the versioning will be checked (e.g. actions/setup-python), but something like mamba-org/provision-with-micromamba --> mamba-org/setup-micromamba will not be picked up.

I'm aware. 

On this note, I have noticed that `actions/create-release` and `actions/upload-release-asset` are both archived, so we should maybe look for another solution. We did something else recently in pyogrio, so might be worth checking that. But not in this PR.

jGaboardi (approved):

PR #518, opened 2023-11-22
RLS: add a changelog for 0.7.0

martinfleis:
@jGaboardi I think that moving forward, we should adopt what the rest of the ecosystem is doing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,15 @@
 Changelog
 =========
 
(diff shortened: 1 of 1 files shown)


jGaboardi (approved):

jGaboardi:
> @jGaboardi I think that moving forward, we should adopt what the rest of the ecosystem is doing.

All good by me.

jGaboardi:
Anything more to add or ready to merge?

PR #522, opened 2023-11-25
support strings in `elements.get_node_id()`

### LINKED ISSUE #519: BUG: get_node_id does not support string IDs
If `edge_id` column is using string IDs, this line will fail as numpy supports only numeric dtypes.

https://github.com/pysal/momepy/blob/78c1b244f3acef60d77cbce98e3426c755b3033d/momepy/elements.py#L807C16-L807C24

We need to change it to `pd.isna`.

jGaboardi:
This PR resolves #519

### DIFF
--- a/momepy/tests/test_elements.py
+++ b/momepy/tests/test_elements.py
@@ -159,12 +159,13 @@ class TestElements:
         assert not ids1.isna().any()
 
         # test for NaNs within `object` nIDs column
+        edges["nID"] = edges["nID"].astype(str)
         _df_buildings = self.df_buildings.copy()
+        _df_buildings["nID"] = _df_buildings["nID"].astype(str)
         _df_buildings.loc[[0, 1], "nID"] = pd.NA
         ids2 = mm.get_node_id(_df_buildings, nodes, edges, "nodeID", "nID")
         assert ids2.isna().sum() == 2
         np.testing.assert_array_equal(ids2[ids2.isna()].index, [0, 1])
-        np.testing.assert_array_equal(ids1.loc[2:], ids2.loc[2:])
 
     def test_get_node_id_ratio(self):
         nx = mm.gdf_to_nx(self.df_streets)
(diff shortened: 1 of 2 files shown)


jGaboardi:
Seeing now that the line isn't being hit by any tests...

martinfleis:
It is technically covering a corner case that should not happen when everything is as expected.

We hit it when we don't have a street network edge attached to a building footprint, which happens only if a building is really far away from any street. But we could possibly simulate that by using `momepy.get_network_id` with a small `min_size` (that API and implementation there are bad...)

jGaboardi:
> It is technically covering a corner case that should not happen when everything is as expected.
> 
> We hit it when we don't have a street network edge attached to a building footprint, which happens only if a building is really far away from any street. But we could possibly simulate that by using `momepy.get_network_id` with a small `min_size` (that API and implementation there are bad...)

Working on creating a synthetic test case now that simply converts some edge IDs in the `bubenec` streets to NaN.

jGaboardi:
Passing locally and hitting the offending line.

momepy/tests/test_elements.py
```diff
@@ -154,8 +155,16 @@ def test_get_node_id(self):
...
             self.df_buildings, self.df_streets, "nID"
         )
-        ids = mm.get_node_id(self.df_buildings, nodes, edges, "nodeID", "nID")
-        assert not ids.isna().any()
+        ids1 = mm.get_node_id(self.df_buildings, nodes, edges, "nodeID", "nID")
+        assert not ids1.isna().any()
+
+        # test for NaNs within `object` nIDs column
+        _df_buildings = self.df_buildings.copy()
+        _df_buildings.loc[[0, 1], "nID"] = pd.NA
```
martinfleis, line 163:
```suggestion
        _df_buildings["nID"] = _df_buildings["nID"].astype(str)
        _df_buildings.loc[[0, 1], "nID"] = pd.NA
```

Might be worth actually testing for the string ID?

jGaboardi: Sure. Will push up a new version.

martinfleis (approved):
Thanks!

PR #532, opened 2024-01-08
ENH: first bits of refactored dimension

martinfleis:
WIP but comes with a lot of simplification over the original API and quite nice speedups compared to original algos.

Using Prague as a benchmark:

```
longest_axis_length
1.59 s vs 3.76 s

perimeter_wall (including contiguity weights creation)
10.8 s vs 29.5 s

perimeter_wall (excluding contiguity weights creation)
4.91 s vs 21.2 s
```

The rest is using roughly the same implementation as before.

@jGaboardi the API change is a bit more radical than I originally anticipated but I believe we should do it properly this time.


martinfleis:
@jGaboardi I am not sure what is the best way of including the functional API in the documentation for the period when both are included. Let's not add it for now and figure it out before releasing? We'll need to rework the user guide as well so we can do that in one go.

jGaboardi:
> Let's not add it for now and figure it out before releasing?

Works for me.

jGaboardi:
>  quite nice speedups compared to original algos.

Very nice!

jGaboardi:
This failure in `dev` is being fixed in #533, correct?

momepy/functional/_dimension.py
```diff
@@ -0,0 +1,144 @@
...
+    Returns
+    -------
+    pandas.Series
+    """
+    return shapely.minimum_bounding_radius(gdf.geometry) * 2
+
+
+def perimeter_wall(gdf, graph=None):
+    """
+    Calculate the perimeter wall length the joined structure.
```
jGaboardi, line 107: Is this the correct docstring? Should it read more like [`Perimeter`](https://github.com/pysal/momepy/blob/2051f4b7aaa353bf28c12f817c2f6f8c7f628ae5/momepy/dimension.py#L65)?

martinfleis: Yes, it is correct. It is reimplementation of [`PerimeterWall`](https://github.com/pysal/momepy/blob/2051f4b7aaa353bf28c12f817c2f6f8c7f628ae5/momepy/dimension.py#L845) not `Perimeter`. `Perimeter` will be removed and users pointed to `.length`.

jGaboardi (approved):

PR #533, opened 2024-01-09
CI/TST: fix test, rm log action, ruff-format

martinfleis:
Fixing the issue in dev CI (that is actually a regression on dev GeoPandas but we can avoid it) and removing the xarray-contrib/issue-from-pytest-log action that does not work.

edit: also resolves #531
edit2: adopt ruff-format

### DIFF
--- a/momepy/preprocessing.py
+++ b/momepy/preprocessing.py
@@ -1113,7 +1113,7 @@ class FaceArtifacts:
                 [gdf.unary_union]
             ).geoms,  # get parts of the collection from polygonize
             crs=gdf.crs,
-        ).explode(ignore_index=True)  # shoudln't be needed but doesn't hurt to ensure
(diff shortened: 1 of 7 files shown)


momepy/preprocessing.py
```diff
@@ -1113,9 +1113,7 @@ def __init__(
                 [gdf.unary_union]
             ).geoms,  # get parts of the collection from polygonize
             crs=gdf.crs,
-        ).explode(
-            ignore_index=True
-        )  # shoudln't be needed but doesn't hurt to ensure
+        ).explode(ignore_index=True)  # shoudln't be needed but doesn't hurt to ensure
```
jGaboardi, line 1116:
```suggestion
        ).explode(ignore_index=True)  # shouldn't be needed but doesn't hurt to ensure
```

jGaboardi (approved):

PR #534, opened 2024-01-12
TYP: add type hints to functional._dimension

martinfleis:
I suppose this is how typing in momepy can look? Is that an improvement? Shall we do it?

### DIFF
--- a/momepy/functional/_dimension.py
+++ b/momepy/functional/_dimension.py
@@ -15,9 +15,9 @@ __all__ = [
 
 
 def volume(
-    area: NDArray[np.float_] | Series[float],
-    height: NDArray[np.float_] | Series[float],
-) -> NDArray[np.float_] | Series[float]:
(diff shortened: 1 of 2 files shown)


momepy/functional/_dimension.py
```diff
@@ -11,7 +14,10 @@
 ]
 
 
-def volume(area, height):
+def volume(
+    area: NDArray[np.float_] | Series,
```
martinfleis, line 18: Maybe we can do with a bare `NDArray`? Since I don't know how to add the same to Series anyway.

jGaboardi: Did you try: `Series[np.float_]`

jGaboardi (approved):
Looks good.

jGaboardi:
Let's merge then. Can always adjust later if we figure it out.

Issue #537, opened 2024-01-19
BUG: tessellation may produce overlapping polygons


martinfleis:
Attached is a tiny portion of large tessellation which contains overlapping polygons. Only very tiny overlaps, but overlaps nevertheless. This causes `shapely.coverage_union_all` to fail. We should ensure that the result of tessellation is a clean topological coverage. 

I guess that some internals of the tessellation algo are in a need of refactoring but I am really not looking forward to that as that code is a beast.

martinfleis:
One pathway to explore is to use GEOS instead of SciPy to generate the diagram. Unsure about performance but shapely gives us back polygons.

martinfleis:
Resolved by #559

PR #538, opened 2024-01-19
ENH: reimplement tessellation using shapely

martinfleis:
WIP but f****ing hell how the time has changed what is possible now.

Performance comparison on 10k buildings -  new`Wall time: 35 s`, old `Wall time: 47.7 s`

Still space for improvements I think. What is great is that GEOS gives us directly polygons which we know are topologically clean.

### DIFF
--- a/momepy/functional/_elements.py
+++ b/momepy/functional/_elements.py
@@ -24,9 +24,13 @@ def tessellation(
     voronoi = shapely.voronoi_polygons(
         shapely.GeometryCollection(objects.values), extend_to=limit
     )
-    geoms = gpd.GeoDataFrame(geometry=shapely.get_parts(voronoi), crs=gdf.crs)
+    geoms = gpd.GeoSeries(shapely.get_parts(voronoi), crs=gdf.crs)
     ids_objects, ids_geoms = geoms.sindex.query(objects, predicate="intersects")
-    polygons = geoms.iloc[ids_geoms].dissolve(objects.index.take(ids_objects))
+    polygons = (
+        geoms.iloc[ids_geoms]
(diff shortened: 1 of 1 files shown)


martinfleis:
Got it down to `Wall time: 10.6 s` in 2ca9832 :)

martinfleis:
Squeezed it under 10s to 9.77, 5x times faster than before with cleaner geometry (it resolves #537). 

It is at the moment less robust in terms of input data. Original tessellation does not error on overlapping polygons and similar issues. It produces erroneous geometry but produces something. This fails on GEOSException. I am not sure which way is better but I tend to say that geometries should be cleaned before running tessellation. We can probably provide some tooling for that, maybe on top of @sjsrey's [`geoplanar`](https://github.com/sjsrey/geoplanar).

Still missing support of LineString tessellation (edge points conflict) and enclosed tessellation.

martinfleis:
I'm thinking that since this is fairly generic polygon-based Voronoi tessellation it may be better to have it in libpysal.cg, no? Rather than hidden in momepy.

martinfleis:
When passing a `limit`, the main bottleneck becomes `clip`. See the profiling below. Might be worth looking in https://github.com/geopandas/geopandas/issues/1803#issuecomment-778637626 but the issue is not union here but a ton of intersections that is essentially void since the whole geom is within the limit. I might need to look into `clip` in geopandas and try to fix this over there.

Screenshot 2024-01-20 at 17 20 58

martinfleis:
Will be superseded by https://github.com/pysal/libpysal/pull/678

martinfleis:
Superseded by https://github.com/pysal/libpysal/pull/678 and #559

PR #545, opened 2024-02-02
gdf2osmnx

anastassiavybornova:
Added the `osmnx_like` keyword to `gdf_to_nx()`. If `osmnx_like=True`, within `_generate_primal`:
* *first*, add all unique node coordinates as nodes to `net` with unique dummy IDs 
*  break each linestring in gdf into two-point linestrings
* for each two-point linestring, find corresponding node IDs (u,v)
* *only then*, add edges as (u,v) pairs to graph _iteratively_, at each step checking whether (u,v) already exists in graph (if so, key+=1)
* the output MultiDiGraph from `gdf_to_nx()`now works as input for `osmnx.simplify_graph()`

### DIFF
--- a/momepy/preprocessing.py
+++ b/momepy/preprocessing.py
@@ -150,9 +150,9 @@ def preprocess(
                         geoms.append(blg[blg["mm_uid"] == j].iloc[0].geometry)
                         blg.drop(blg[blg["mm_uid"] == j].index[0], inplace=True)
                 new_geom = shapely.ops.unary_union(geoms)
-                blg.loc[
-                    blg.loc[blg["mm_uid"] == key].index[0], blg.geometry.name
-                ] = new_geom
+                blg.loc[blg.loc[blg["mm_uid"] == key].index[0], blg.geometry.name] = (
+                    new_geom
+                )
 
         blg.drop(delete, inplace=True)
     return blg[buildings.columns]
@@ -1568,7 +1568,9 @@ class FaceArtifacts:
                 [gdf.unary_union]
             ).geoms,  # get parts of the collection from polygonize
             crs=gdf.crs,
-        ).explode(ignore_index=True)  # shouldn't be needed but doesn't hurt to ensure
+        ).explode(
+            ignore_index=True
+        )  # shouldn't be needed but doesn't hurt to ensure
 
         # Store geometries as a GeoDataFrame
         self.polygons = gpd.GeoDataFrame(geometry=polygons)
--- a/momepy/utils.py
+++ b/momepy/utils.py
@@ -8,7 +8,8 @@ import libpysal
 import networkx as nx
 import numpy as np
 from numpy.lib import NumpyVersion
-from shapely.geometry import Point
+from shapely.geometry import Point, LineString
+from shapely.ops import linemerge
 
 __all__ = [
     "unique_id",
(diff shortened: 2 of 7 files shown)


momepy/preprocessing.py
```diff
@@ -150,9 +150,9 @@ def preprocess(
                         geoms.append(blg[blg["mm_uid"] == j].iloc[0].geometry)
                         blg.drop(blg[blg["mm_uid"] == j].index[0], inplace=True)
                 new_geom = shapely.ops.unary_union(geoms)
-                blg.loc[
-                    blg.loc[blg["mm_uid"] == key].index[0], blg.geometry.name
-                ] = new_geom
+                blg.loc[blg.loc[blg["mm_uid"] == key].index[0], blg.geometry.name] = (
```
jGaboardi, line 153: Is this extra indexing needed due to an update in Pandas/GeoPandas?

momepy/utils.py
```diff
@@ -47,7 +48,14 @@ def _angle(a, b, c):
     return abs((a2 - a1 + 180) % 360 - 180)
 
 
-def _generate_primal(graph, gdf_network, fields, multigraph, oneway_column=None):
+def _make_edgetuples(nodelist):
+    """Zip list of nodes into list of edgetuples. Helper function for _generate_primal"""
+    return [z for z in zip(nodelist, nodelist[1:])]
```
jGaboardi, line 53: Can we make this more robust with?:

```suggestion
    return [z for z in zip(nodelist[:-1], nodelist[1:], strict=True)]
```

momepy/utils.py
```diff
@@ -71,24 +79,67 @@ def _generate_primal(graph, gdf_network, fields, multigraph, oneway_column=None)
...
-            key += 1
-
-            if oneway_column:
-                oneway = bool(getattr(row, oneway_column))
-                if not oneway:
-                    graph.add_edge(last, first, key=key, **attributes)
-                    key += 1
-        else:
-            graph.add_edge(first, last, **attributes)
+        gdf_network["geometry"] = gdf_network["geometry"].apply(
```
jGaboardi, line 84: If this will always be a GeoDatFrame, should we use the `geometry` attribute here?
```suggestion
        gdf_network.geometry = gdf_network.geometry.apply(
```

jGaboardi (commented):
Looking very good @anastassiavybornova.

> * break each linestring in gdf into two-point linestrings
> * for each two-point linestring, find corresponding node IDs (u,v)

I am curious, does this approach handle loops (e.g., $[u,u]$ one-point linestrings)? Or does "two-point linestrings" simply imply both endpoints of the linestring (in which case ignore my question)?

martinfleis (commented):
Did only a partial review as I need to clarify some bits.

What exactly is the difference between the two options? Assuming both are MultiDiGraphs. As far as I can tell, the only difference is that the original uses tuple `(x, y)` to encode node with no attributes, the osmnx-like encodes nodes to integers and adds x and y as attributes. Is there anything else I don't see?

If I am right, then my suggestion would be to adapt our original function to add `{'x': 1, 'y': 2}` attributes to each node as osmnx-like version do and if osmnx_like=True, call [`convert_node_labels_to_integers`](https://networkx.org/documentation/stable/reference/generated/networkx.relabel.convert_node_labels_to_integers.html#networkx.relabel.convert_node_labels_to_integers) and be done with it. Am I wrong?

martinfleis:
Looking into it a bit further, it seems that the only change we need to do is to add the attributes with coordinates to nodes. Adding these two lines results in no issue on a small sample dataset.

```py
node_attrs = {node: {"x": node[0], "y": node[1]} for node in graph.nodes}
nx.set_node_attributes(graph, node_attrs)
```

momepy/preprocessing.py
anastassiavybornova: haven't touched that one (except for running black on it) i'll leave the question to @martinfleis

martinfleis:
@anastassiavybornova maybe have a look at #546 before spending more time on this.

momepy/utils.py
```diff
@@ -71,24 +79,67 @@ def _generate_primal(graph, gdf_network, fields, multigraph, oneway_column=None)
...
+            lambda x: linemerge(x) if x.geom_type == "MultiLineString" else x
```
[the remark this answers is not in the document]
anastassiavybornova: _generate_primal already checks for some potential non-linestring geometries, so i thought i'd add this one as well - but yes it is a pre-processing step indeed. fine to leave it out of the function.

PR #553, opened 2024-02-29
ENH: add orientation and shared_walls functional versions

martinfleis:
Orientation:
```
    # buildings
    # old: 3.23 s ± 77 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
    # new: 932 ms ± 12.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
    # streets
    # old: 374 ms ± 5.56 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
    # new: 62 ms ± 681 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Shared walls:

```
    # 8.68 s ± 28.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
    # 10.3 s ± 17.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

This is slower but it does not cause erroneous results coming from overlapping polygons.

### DIFF
--- a/momepy/__init__.py
+++ b/momepy/__init__.py
@@ -8,6 +8,7 @@ from .distribution import *
 from .diversity import *
 from .elements import *
 from .functional._dimension import *
+from .functional._distribution import *
 from .functional._shape import *
 from .graph import *
 from .intensity import *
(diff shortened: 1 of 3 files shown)


jGaboardi (approved):
LGTM. One small typo.

PR #554, opened 2024-03-06
ENH: add alignment to the new API

martinfleis:
Functional version of alignment based on libpysal Graph. It is fast. 282x faster than the original.

```
 61.6 ms ± 3.63 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
 17.4 s ± 1.13 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
```


jGaboardi (approved):
LGTM

PR #556, opened 2024-03-08
ENH: mean_interbuilding_distance and building_adjacency

martinfleis:
Gosh I spent way too much time on this...

mean_interbuilding_distance:

```
35s new
57s old
```

Given my first attempt took 5mins, I call this a victory. 

building_adjacency:

```
 old: 251 ms ± 14.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
 new: 422 ms ± 12.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

of which 330 ms is `assign_self_weight` which will be used in other functions so I'm wondering if it is worth adding a keyword `self_weighted=True` to skip this. In practice it is nearly negligible difference though...

These two functions also quite significantly change the API compared to the original. @jGaboardi can you give it a thought whether it makes sense in this way? I think it does and it is more flexible than before but it is different.

### DIFF
--- a/momepy/functional/_distribution.py
+++ b/momepy/functional/_distribution.py
@@ -1,7 +1,4 @@
-import warnings
-
 import geopandas as gpd
-import networkx as nx
 import numpy as np
 import pandas as pd
 import shapely
@@ -9,6 +6,7 @@ from geopandas import GeoDataFrame, GeoSeries
 from libpysal.graph import Graph
 from packaging.version import Version
 from pandas import Series
+from scipy import sparse
 
 __all__ = [
     "orientation",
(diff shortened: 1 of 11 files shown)


momepy/functional/_distribution.py
```diff
@@ -152,3 +161,112 @@ def neighbor_distance(geometry: GeoDataFrame | GeoSeries, graph: Graph) -> Serie
...
+        Graph representing the extent around each geometry within which to calculate
+        the level of building adjacency. This can be a distance based graph, KNN graph,
+        higher order contiguity, etc.
+
+    Returns
+    -------
+    Series
+    """
+    components = contiguity_graph.component_labels
+    neighborhood_graph = neighborhood_graph.assign_self_weight()
```
jGaboardi, line 264: [`libpysal==4.8`](https://github.com/pysal/momepy/actions/runs/8209874874/job/22456245168?pr=556#step:3:649) did not have [`Graph().assign_self_weight()`](https://github.com/pysal/libpysal/blob/v4.8.0/libpysal/graph/base.py), which is causing [the failure here](https://github.com/pysal/momepy/actions/runs/8209874874/job/22456245168?pr=556#step:5:429).

jGaboardi:
> These two functions also quite significantly change the API compared to the original. @jGaboardi can you give it a thought whether it makes sense in this way? I think it does and it is more flexible than before but it is different.

I think it makes sense, cleaner and more efficient for sure.

momepy/functional/_distribution.py
jGaboardi: Were we thinking a skip on `mean_interbuilding_distance()` for now or build in logic for use with `libpysal<4.10`?

martinfleis: I would just bump min requirement.

martinfleis:
> of which 330 ms is assign_self_weight which will be used in other functions so I'm wondering if it is worth adding a keyword self_weighted=True to skip this. In practice it is nearly negligible difference though...

Solved in 0d897cc by cheaply (10ms) checking if the graph already has self-weights and assigning only if they do not. Which means that if you pass self-weighted graph, the new implementation takes about half the time compared to the original.

jGaboardi (approved):

PR #557, opened 2024-03-21
ENH: add neighbors

martinfleis:
Easy one this time. Hard to time, as `cardinalities` are cached.

```
# first run (weighted)
    # CPU times: user 49.5 ms, sys: 121 ms, total: 170 ms
    # Wall time: 179 ms
# timeit
    # 25.7 ms ± 190 µs per loop (mean ± std. dev. of 7 runs, 10 loops each) - new
    # 549 ms ± 2.07 ms per loop (mean ± std. dev. of 7 runs, 1 loop each) - old
```

 Faster esp. if other function before called cardinalities.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,7 +56,7 @@ momepy = ["datasets/bubenec.gpkg", "datasets/tests.gpkg"]
 [tool.ruff]
 line-length = 88
 select = ["E", "F", "W", "I", "UP", "N", "B", "A", "C4", "SIM", "ARG"]
-ignore = ["B006", "F403"]
+ignore = ["B006", "F403", "SIM108"]
 exclude = ["momepy/tests/*", "docs/*", "benchmarks/*"]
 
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -56,7 +56,7 @@ momepy = ["datasets/bubenec.gpkg", "datasets/tests.gpkg"]
 [tool.ruff]
 line-length = 88
 select = ["E", "F", "W", "I", "UP", "N", "B", "A", "C4", "SIM", "ARG"]
-ignore = ["B006", "F403"]
+ignore = ["B006", "F403", "SIM108"]
```
martinfleis, line 59: This is to avoid converting

```py
    if weighted:
        r = graph.cardinalities / geometry.length
    else:
        r = graph.cardinalities
```
into an inline. I find this way more readable.

jGaboardi (approved):

PR #561, opened 2024-03-24
GHA: move user guide testing to separate action, add workflow dispatch

### LINKED ISSUE #560: Test user guide only on main?
Currently, we execute notebooks in the user guide to test them for failure on every CI run, including PRs. Do we want to keep doing it? It might be enough to run that only on main and in the nightly CI. @jGaboardi thoughts? I takes 11 minutes.

martinfleis:
Closes #560


jGaboardi (approved):
LGTM

PR #562, opened 2024-03-24
test efficiency of graphblas in straightness

martinfleis:
This is more of an exploration than anything. I am curious if `graphblas_algorithms` engine for networkx can help us with our stuff, so I reimplemented `straigtness_centrality` on top of it. For now only the core part. Using the network from our [docs](http://docs.momepy.org/en/stable/user_guide/graph/centrality.html), I am able to get it from 1min 50s down to 17s, which means it is 6.5x faster. There's nothing to shave from our overhead as 98% of time is now spent on `all_pairs_bellman_ford_path_length`. I was expecting a bit more from this, given the speedups in benchmarks in their Readme are in hundreds, but better than nothing :). 

The code for networkx and graphblas is similar but not the same. Networkx-compatible code is possible but much slower than this. So having graphblas as optional dependency (like numba elsewhere in PySAL) would mean having two implementations, which I am not a fan of. But as @jGaboardi pointed out, graphblas is not installable on Apple Silicon under Python 3.12 at the moment. If there'll be delays of this sort, we I am not sure I like depending on it. (Though it is installable from PyPI...).

### DIFF
--- a/momepy/graph.py
+++ b/momepy/graph.py
@@ -3,7 +3,6 @@
 # connectivity.py
 # definitions of connectivity characters
 import collections
-import math
 
 import networkx as nx
 import numpy as np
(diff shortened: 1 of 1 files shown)


martinfleis:
On the installation, python-graphblas switched from noarch to compiled code some time ago but conda-forge is not set to build arm64. Fixing that in https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/5676. 

Once the recipe gets rebuilt, the installation should not be an issue any longer.

martinfleis:
I'll close it as it was a test. It seems that while the performance can be improved, `graphblas_algorithms` might need to mature a bit before relying on it.

Issue #564, opened 2024-04-02
CI failure – `TestElements.test_Blocks_non_default_index`


jGaboardi:
* [CI failure](https://github.com/pysal/momepy/actions/runs/8514089728/job/23319088003#step:5:194) – `TestElements.test_Blocks_non_default_index`
* seems to be relating to `pandas`

```
self = <momepy.tests.test_elements.TestElements object at 0x7f103b9c1b80>

    def test_Blocks_non_default_index(self):
        tessellation = self.df_tessellation.copy()
        tessellation.index = tessellation.index * 3
        buildings = self.df_buildings.copy()
        buildings.index = buildings.index * 5
    
        blocks = mm.Blocks(tessellation, self.df_streets, buildings, "bID", "uID")
    
        assert_index_equal(tessellation.index, blocks.tessellation_id.index)
>       assert_index_equal(buildings.index, blocks.buildings_id.index)
E       AssertionError: Index are different
E       
E       Index values are different (100.0 %)
E       [left]:  RangeIndex(start=0, stop=720, step=5)
E       [right]: Index([655, 635, 660, 640, 650, 630, 645, 625, 570, 665,
E              ...
E              405, 365, 610, 615, 360, 355, 350, 345, 340, 335],
E             dtype='int64', length=144)

momepy/tests/test_elements.py:134: AssertionError
```

martinfleis:
I think it is caused by https://github.com/geopandas/geopandas/pull/2353 but I am not sure if it is expected or a regression yet. Will need to have a proper look.

martinfleis:
Upstream issue -> https://github.com/geopandas/geopandas/issues/3239

jGaboardi:
Let's close out here or keep open until actually fixed?

PR #565, opened 2024-04-02
TST: check the degree of error in #564

martinfleis:
just to test whether the values are botched in #564 or the order only.


martinfleis:
Will be fixed in upstream/

PR #566, opened 2024-04-15
ENH: street_alignment and get_nearest_street

martinfleis:
I love how streamlined these things are now compared to the original :).

### DIFF
--- a/momepy/functional/_distribution.py
+++ b/momepy/functional/_distribution.py
@@ -318,18 +318,17 @@ def neighbors(
 
 def street_alignment(
     building_orientation: Series,
-    stret_orientation: Series,
+    street_orientation: Series,
     street_index: Series,
 ) -> Series:
     """Calulate the deviation of the building orientation from the street orientation.
-
 
     Parameters
     ----------
     building_orientation : Series
         Series with the orientation of buildings. Can be measured using
         :func:`orientation`.
-    stret_orientation : Series
+    street_orientation : Series
         Series with the orientation of streets. Can be measured using
         :func:`orientation`.
     street_index : Series
--- a/momepy/functional/_distribution.py
+++ b/momepy/functional/_distribution.py
@@ -339,4 +339,4 @@ def street_alignment(
     -------
     Series
     """
-    return (building_orientation - stret_orientation.loc[street_index].values).abs()
+    return (building_orientation - street_orientation.loc[street_index].values).abs()
--- a/momepy/functional/_distribution.py
+++ b/momepy/functional/_distribution.py
@@ -16,6 +16,7 @@ __all__ = [
     "mean_interbuilding_distance",
     "building_adjacency",
     "neighbors",
+    "street_alignment",
 ]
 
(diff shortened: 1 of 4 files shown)


momepy/functional/_distribution.py
```diff
@@ -313,3 +314,30 @@ def neighbors(
...
     return r
+
+
+def street_alignment(
+    building_orientation: Series,
+    stret_orientation: Series,
+    street_index: Series,
+) -> Series:
+    """Calulate the deviation of the building orientation from the street orientation.
+
```
jGaboardi, line 325:
```suggestion
```

momepy/functional/_distribution.py
```diff
@@ -313,3 +314,30 @@ def neighbors(
 
     r.name = "neighbors"
     return r
+
+
+def street_alignment(
+    building_orientation: Series,
+    stret_orientation: Series,
```
jGaboardi, line 321:
```suggestion
    street_orientation: Series,
```

momepy/functional/_distribution.py
```diff
@@ -313,3 +314,30 @@ def neighbors(
...
+) -> Series:
+    """Calulate the deviation of the building orientation from the street orientation.
+
+
+    Parameters
+    ----------
+    building_orientation : Series
+        Series with the orientation of buildings. Can be measured using
+        :func:`orientation`.
+    stret_orientation : Series
```
jGaboardi, line 332:
```suggestion
    street_orientation : Series
```

momepy/functional/_elements.py
```diff
@@ -295,3 +296,44 @@ def verify_tessellation(tesselation, geometry):
             stacklevel=2,
         )
     return collapsed, multipolygons
+
+
+def get_nearest_street(
+    buildings: gpd.GeoSeries | gpd.GeoDataFrame,
```
jGaboardi, line 302: Is there is specific reasoning to using both `pd` and `Series` above, and using `gpd.GeoSeries` here? Should we also do `from gpd import GeoSeries` to conform? Or vice-versa?

martinfleis: No particular reason. My thought was that people will be familiar with Series but may not be with GeoSeries, so I wanted to indicate where it comes from. Do you have any preference here?

jGaboardi: My personal preference is "always explicit" so if I'd had my way it would be non-aliased & full usage: e.g.:

```python
import pandas
...
pandas.Series
```

However, I know that my preference is not norm. If I had to choose one of the two methods already being used here, I'd probably go with non-aliased classes (e.g. `from gpd import GeoSeries`) for the type hinting.

martinfleis: I also know that you like linting rules, so have one for you - https://docs.astral.sh/ruff/rules/unconventional-import-alias/

martinfleis:
I'll do the typing consistency as a follow-up across the codebase.

jGaboardi (approved):

PR #567, opened 2024-04-15
TYP: consistency in typing

martinfleis:
Series, GeoSeries and GeoDataFrame are imported directly. It was mostly like this but it was inconsistent.

### DIFF
--- a/momepy/functional/_elements.py
+++ b/momepy/functional/_elements.py
@@ -4,6 +4,7 @@ import geopandas as gpd
 import numpy as np
 import pandas as pd
 import shapely
+from geopandas import GeoDataFrame, GeoSeries
 from joblib import Parallel, delayed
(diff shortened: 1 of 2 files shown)


jGaboardi (approved):

momepy/functional/_elements.py
```diff
@@ -238,7 +235,7 @@ def _tess(ix, poly, blg, threshold, shrink, segment, enclosure_id):
         tess[enclosure_id] = ix
         return tess
 
-    return gpd.GeoDataFrame(
+    return GeoDataFrame(
```
jGaboardi, line 238: huh. not currently covered?

martinfleis: Hmm, yeah. Follow-up will come.

PR #568, opened 2024-04-16
BUG: fix enclosed tessellation index allocation when threshold eliminates buildings

martinfleis:
That uncovered line @jGaboardi noticed actually contained a bug...

It is a corner case but when there are at least two intersecting buildings in an enclosure but both of them are eliminated by the intersection threshold, we were getting multiple rows indexed -1. Fixed.


jGaboardi (approved):
Minor typo, but LGTM.

jGaboardi (approved):

PR #570, opened 2024-04-23
ENH: describe as a replacement of AverageCharacter

martinfleis:
I started refactoring `AverageCharacter` and changed my mind a couple of times on what to do about it. In the end, decided to generalise it into `momepy.describe` which computes `"mean", "median", "std", "min", "max", "sum"` and optionally `"mode"` of values within the neighborhood, possibly after cutting the extremes off.  We still have the full functionality of the original `AverageCharacter` folded in though. 

I am still uncertain whether this is the right API, so I'd appreciate some input @jGaboardi and @u3ks. 

Also, this may be more fitting in `libpysal.graph` but it may be wise to keep the API there a bit leaner.

Tests are missing as of now but will add them tomorrow.


momepy/functional/_distribution.py
```diff
@@ -339,3 +343,79 @@ def street_alignment(
...
+
+    if not isinstance(values, Series):
+        values = Series(values)
+
+    grouper = values.take(graph._adjacency.index.codes[1]).groupby(
+        graph._adjacency.index.codes[0]
+    )
+
+    if q is None:
+        stat_ = grouper.agg(["mean", "median", "std", "min", "max", "sum"])
```
martinfleis, line 410: This is super fast if we don't need mode or quantile cut, so the reason to keep this list customisable for performance reasons is gone.

momepy/functional/_distribution.py
```diff
@@ -339,3 +343,79 @@ def street_alignment(
     Series
     """
     return (building_orientation - street_orientation.loc[street_index].values).abs()
+
+
+def describe(
+    values: NDArray[np.float_] | Series,
```
jGaboardi, line 349: Seems to me like many of the mainstream graph-based packages generally have graph-like object as the first positional argument. How would we feel about swapping `values` & `graph` here?

jGaboardi (commented):
Other than the potential swapping of `values` and `graph` I think this looks pretty nice and lean.

momepy/tests/test_utils.py
```diff
@@ -196,9 +196,9 @@ def test_nx_to_gdf_osmnx(self):
         assert len(lines) == 16
 
     def test_limit_range(self):
-        assert list(mm.limit_range(range(10), rng=(25, 75))) == [2, 3, 4, 5, 6, 7]
-        assert list(mm.limit_range(range(10), rng=(10, 90))) == [1, 2, 3, 4, 5, 6, 7, 8]
-        assert list(mm.limit_range([0, 1], rng=(25, 75))) == [0, 1]
+        assert list(mm.limit_range(np.arange(10), rng=(25, 75))) == [2, 3, 4, 5, 6, 7]
```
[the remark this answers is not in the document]
martinfleis: Yeah, it gives an array. limit_range now need an array, simple range does not work.

momepy/functional/_distribution.py
martinfleis: So far, we've been using this order. In alignment we have `(orientation, graph)`, in neighbor_distance we have `(geometry, graph)`, mean_interbuilding_distance has `(geometry, adjacency_graph, neighborhood_graph)`. I don't mind much in which order it is gonna be as long as we can make it consistent.

What we have here more mirrors what we have elsewhere in pysal, like esda's `Geary(y, w)`. (which also reminds me that we should probably rename `values` to `y` or `a` or something that is common in the ecosystem.

martinfleis:
Some timings:

```
no mode, no quantiles
# new: 141 ms ± 626 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
# old: Wall time: 12.7 s

with mode, no quantiles
# new Wall time: 15.1 s
# old Wall time: 30.2 s

no mode, quantiles
# new Wall time: 8.54 s
# old Wall time: 14.4 s

with mode and quantiles
# new Wall time: 23.6 s
# old Wall time: 30.8 s
```
So we're faster in all cases. How much depends on the use case.

momepy/functional/_distribution.py
jGaboardi: > So far, we've been using this order

Let's keep the order how it is then.

momepy/functional/tests/_testing.py
jGaboardi, line 1: Shall we rename this to `conftest.py`? Then we can start throwing fixtures in here as needed, etc. plus any other nifty streamlining via `pytest`.

Also, we can turn `assert_result()` into a parameterized fixture, like over in `spopt` (fixture [here](https://github.com/pysal/spopt/blob/785b396d8459491fee6e5a826b02b513fabf89f3/spopt/tests/conftest.py#L23), example call [here](https://github.com/pysal/spopt/blob/785b396d8459491fee6e5a826b02b513fabf89f3/spopt/tests/test_lscp.py#L29)) if desired. Though it's tad Frankenstein-ish; may be better to keep as a helper func.

jGaboardi: > > Also, we can turn assert_result() into a parameterized fixture
> 
> why?

Just if we wanted to have it as an actual fixture. But def not really needed here as long as we can have it as a reusable function.

martinfleis: I'd keep it as is.

jGaboardi (approved):

martinfleis:
Merging but have numba version of the same in works. Will do follow-up.

PR #575, opened 2024-04-29
functional reached calculations

u3ks:
a new functional version of the mm.Reached function. It calculates several statistics at once similar to mm.describe(

### DIFF
--- a/momepy/functional/_diversity.py
+++ b/momepy/functional/_diversity.py
@@ -7,11 +7,12 @@ from numpy.typing import NDArray
 from packaging.version import Version
 from pandas import DataFrame, Series
 
-throw_numba_warning = False
 try:
     from numba import njit
+
+    HAS_NUMBA = True
 except (ModuleNotFoundError, ImportError):
-    throw_numba_warning = True
+    HAS_NUMBA = False
     from libpysal.common import jit as njit
 
 __all__ = ["describe", "describe_reached"]
@@ -36,11 +37,22 @@ def _describe(values, q, include_mode=False):
     """Helper function to calculate average."""
     nan_tracker = np.isnan(values)
 
-    if (len(values) > 2) and (not nan_tracker.all()):
-        if nan_tracker.any():
-            lower, higher = np.nanpercentile(values, q)
-        else:
-            lower, higher = np.percentile(values, q)
+    if nan_tracker.all():
+        return [
+            np.nan,
+            np.nan,
+            np.nan,
+            np.nan,
+            np.nan,
+            np.nan,
+            np.nan,
+        ]
+
+    else:
+        values = values[~np.isnan(values)]
+
+    if len(values) > 2:
+        lower, higher = np.percentile(values, q)
         values = values[(lower <= values) & (values <= higher)]
 
     results = [
@@ -52,8 +64,10 @@ def _describe(values, q, include_mode=False):
         np.max(values),
         np.sum(values),
     ]
+
     if include_mode:
         results.append(_mode(values))
+
     return results
 
 
@@ -135,7 +149,7 @@ def describe(
         A DataFrame with descriptive statistics.
     """
 
-    if throw_numba_warning:
+    if not HAS_NUMBA:
         warnings.warn(
             "The numba package is used extensively in this function to accelerate the"
             " computation of statistics but it is not installed or  cannot be imported."
@@ -199,7 +213,7 @@ def describe_reached(
     if Version(pd.__version__) <= Version("2.1.0"):
         raise NotImplementedError("Please update to a newer version of pandas.")
 
-    if throw_numba_warning:
+    if not HAS_NUMBA:
         warnings.warn(
             "The numba package is used extensively in this function to accelerate the"
             " computation of statistics but it is not installed or  cannot be imported."
--- a/momepy/functional/_diversity.py
+++ b/momepy/functional/_diversity.py
@@ -1,11 +1,101 @@
 import warnings
 
 import numpy as np
+import pandas as pd
 from libpysal.graph import Graph
 from numpy.typing import NDArray
 from pandas import DataFrame, Series
 
-__all__ = ["describe"]
+try:
+    from numba import njit
+except (ModuleNotFoundError, ImportError):
+    warnings.warn(
+        "The numba package is used extensively in this function to accelerate the"
+        " computation of statistics but it is not installed or  cannot be imported."
+        " Without numba, these computations may become slow on large data.",
+        UserWarning,
+        stacklevel=2,
+    )
+    from libpysal.common import jit as njit
+
+__all__ = ["describe", "describe_reached"]
+
+
+@njit
+def _mode(array):
+    """Custom mode function for numba."""
+    array = np.sort(array.ravel())
+    mask = np.empty(array.shape, dtype=np.bool_)
+    mask[:1] = True
+    mask[1:] = array[1:] != array[:-1]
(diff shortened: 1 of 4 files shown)


momepy/functional/_diversity.py
```diff
@@ -1,11 +1,101 @@
...
+    from numba import njit
+except (ModuleNotFoundError, ImportError):
+    warnings.warn(
+        "The numba package is used extensively in this function to accelerate the"
+        " computation of statistics but it is not installed or  cannot be imported."
+        " Without numba, these computations may become slow on large data.",
+        UserWarning,
+        stacklevel=2,
+    )
+    from libpysal.common import jit as njit
```
martinfleis, line 19: The warning should be printed only if a user calls a function that actually uses numba. That is why it was originally inside `describe`.  I would still keep the warning inside if possible.

momepy/functional/_diversity.py
```diff
@@ -48,72 +138,103 @@ def describe(
...
 
-    if q is None:
-        stat_ = grouper.agg(["mean", "median", "std", "min", "max", "sum"])
-        if include_mode:
-            stat_["mode"] = grouper.agg(lambda x: _mode(x.values))
+    return _compute_stats(grouper, q, include_mode)
+
+
+def describe_reached(
+    y, y_group, result_index=None, graph=None, q=None, include_mode=False
```
martinfleis, line 152: I would probably call `y_group` `graph_index` following what we did in `street_alignment`
https://github.com/pysal/momepy/blob/36b13f827200c694ecbcff54545b227d526959b6/momepy/functional/_distribution.py#L318-L322

it should be clear from the name that this contains ID's aligned with `y` that link to the index in `graph`.

momepy/functional/_diversity.py
```diff
@@ -48,72 +138,103 @@ def describe(
...
+) -> DataFrame:
+    """
+    Calculates statistics of ``y`` objects reached on a street network.
```
martinfleis, line 155: Not only street network. Graph or IDs can come from anywhere. You can easily use enclosure ID here instead. The function is pretty generic.

momepy/functional/_diversity.py
```diff
@@ -48,72 +138,103 @@ def describe(
...
+    Requires a ``y_group`` that links the ``y`` objects to streets (or ``graph``)
+    assigned beforehand (e.g. using :py:func:`momepy.get_network_id`).
```
martinfleis, line 157:
```suggestion
    assigned beforehand (e.g. using :py:func:`momepy.get_nearest_street`).
```

that is the new one

momepy/functional/_diversity.py
```diff
@@ -48,72 +138,103 @@ def describe(
...
-            cols.append("mode")
-        stat_.columns = cols
+        raise ValueError(
+            "One of result_index or graph has to be specified, but not both."
+        )
 
-    return stat_
+    if isinstance(y, np.ndarray):
+        y = pd.Series(y, name="obs_index")
+    elif isinstance(y, Series):
```
martinfleis, line 208:
```suggestion
    if isinstance(y, Series):
```
to treat the series created above, no?

momepy/functional/_diversity.py
```diff
@@ -48,72 +138,103 @@ def describe(
...
+        )
+
+    stats = _compute_stats(grouper, q, include_mode)
+
+    result = pd.DataFrame(
+        np.zeros((result_index.shape[0], stats.shape[1])), index=result_index
+    )
+    result.loc[stats.index.values] = stats.values
+    result.columns = stats.columns
+    result = result.fillna(0)
```
martinfleis, line 238: No need for this if you are filling the results full of zeros, no?

momepy/functional/tests/test_diversity.py
```diff
@@ -11,7 +12,7 @@
 GPD_013 = Version(gpd.__version__) >= Version("0.13")
 
 
-class TestDistribution:
+class TestDiscribe:
```
martinfleis, line 15:
```suggestion
class TestDescribe:
```

typo

momepy/functional/tests/test_diversity.py
```diff
@@ -115,3 +116,141 @@ def test_describe_array(self):
...
+            )
+
+    def test_describe_reached(self):
+        df = mm.describe_reached(
+            self.df_buildings[["area", "fl_area"]],
+            self.df_buildings["nID"],
+            self.df_streets.index,
+        )
+        # not using assert_result since the method
+        # is returning an aggregation, indexed based on nID
```
martinfleis, line 172: Why not? We use it in `describe` without issues. I'd prefer to use it as it checks the index.

martinfleis (commented):
First round

momepy/functional/_diversity.py
```diff
@@ -1,11 +1,97 @@
...
+        values = values[(lower <= values) & (values <= higher)]
+
+    results = [
+        values.shape[0],
+        np.mean(values),
+        np.median(values),
+        np.std(values),
+        np.min(values),
+        np.max(values),
+        np.sum(values),
```
martinfleis, line 53: Needs to work properly with missing values.

u3ks: I updated the code, I think we can just get rid of the nans beforehand and compute everything on the non_na values, what do you think?

momepy/functional/_diversity.py
```diff
@@ -1,11 +1,111 @@
...
+
+def _compute_stats(grouper, q=None, include_mode=False):
+    """
+    Fast compute of "count", "mean", "median", "std", "min", "max", "sum" statistics,
+    with optional mode and quartile limits.
+
+    Parameters
+    ----------
+    grouper : pandas.GroupBy
+        Groupby Object which specifies the aggregations to be performed
```
jGaboardi, line 82:
```suggestion
        Groupby Object which specifies the aggregations to be performed.
```

momepy/functional/_diversity.py
```diff
@@ -104,16 +165,112 @@ def _describe(values, q, include_mode=False):
...
+        Compute mode along with other statistics. Default is False. Mode is
+        computationally expensive and not useful for continous variables.
+
+    Returns
+    -------
+    DataFrame
+    """
+
+    if Version(pd.__version__) <= Version("2.1.0"):
+        raise NotImplementedError("Please update to a newer version of pandas.")
```
jGaboardi, line 214: Is this a [proper usage](https://stackoverflow.com/a/44316063/8692996) of a `NotImplementedError`?

momepy/functional/_diversity.py
```diff
@@ -104,16 +165,112 @@ def _describe(values, q, include_mode=False):
...
+            "The numba package is used extensively in this function to accelerate the"
+            " computation of statistics but it is not installed or  cannot be imported."
+            " Without numba, these computations may become slow on large data.",
+            UserWarning,
+            stacklevel=2,
+        )
+
+    if (result_index is None) and (graph is None):
+        raise ValueError(
+            "One of result_index or graph has to be specified, but not both."
```
jGaboardi, line 227: Let's declare this `ValueError` so it can be used here and below. Or at least declare the message to re-use.

momepy/functional/_diversity.py
martinfleis: should be `ImportError` I think

martinfleis (commented):
Few notes, mostly on the documentation side.

martinfleis (approved):

PR #576, opened 2024-05-02
functional node_density implementation

u3ks:
functional implementation of nodedensity.
I kept the function logic pretty much the same, just used the new graph api and np.in1d, instead of np.isin.

Timings:
old:
CPU times: user 14.9 s, sys: 69.3 ms, total: 15 s
Wall time: 14.9 s

new:
CPU times: user 6.19 s, sys: 0 ns, total: 6.19 s
Wall time: 6.19 s

Potentially, the function can be sped up further if we change the function API and pass the networkx.MultiGraph object.
Then, we can do:

sum(l for (_,_,l) in nx.subgraph(neighbours).edges.data('mm_len'))

where neighbours comes from the new libpysal.Graph.


momepy/functional/_intensity.py
```diff
@@ -48,3 +50,62 @@ def _calculate_courtyards(group):
     )
 
     return result
+
+
+def node_density(left, right, graph, weighted=False) -> Series:
+    """
+    Calculate the density of nodes neighbours on street network defined in
+    ``spatial_weights``. Calculated as the number of neighbouring
```
martinfleis, line 58:
```suggestion
    """Calculate the density of nodes neighbours on street network defined in
    ``spatial_weights``. 
    
    Calculated as the number of neighbouring
```

momepy/functional/_intensity.py
```diff
@@ -48,3 +50,62 @@ def _calculate_courtyards(group):
...
+    Series
+        A Series containing resulting values.
+
+    Examples
+    --------
+    >>> nodes['density'] = mm.node_density(nodes, edges, sw)
+    """
+
+    if (
+        np.intersect1d(["node_start", "node_end", "mm_len"], right.columns).shape[0]
```
martinfleis, line 88: Don't require `mm_len`, just call `.length` yourself. That is cheap.

martinfleis (commented):
> Potentially, the function can be sped up further if we change the function API and pass the networkx.MultiGraph object.

Can you elaborate? You would then pass the networkx graph and libpysal graph only? I suppose that we can change that as you most likely have the nx graph anyway and if not, it is cheap to generate.

momepy/functional/_intensity.py
```diff
@@ -48,3 +50,67 @@ def _calculate_courtyards(group):
...
+
+    required_cols = ["node_start", "node_end"]
+    if not np.isin(required_cols, right.columns).all():
+        raise ValueError(
+            f"Columns { *required_cols, } are needed in "
+            "right GeoDataframe for the calculations."
+        )
+    if weighted and ("degree" not in left.columns):
+        raise ValueError(
+            "Degree column is needed in " "left GeoDataframe for density calculations."
```
martinfleis, line 99: This has weird formatting. We should also be explicit by saying `"degree"` rather than `Degree`.

momepy/functional/_intensity.py
```diff
@@ -48,3 +50,67 @@ def _calculate_courtyards(group):
...
+    Series
+        A Series containing resulting values.
+
+    Examples
+    --------
+    >>> nodes['density'] = mm.node_density(nodes, edges, graph)
+    """
+
+    required_cols = ["node_start", "node_end"]
+    if not np.isin(required_cols, right.columns).all():
```
martinfleis, line 92: Can we test these errors? With `pytest.raises`?

martinfleis (commented):
@jGaboardi we've been discussing this in person, sorry for not so transparent process!

martinfleis (approved):

momepy/functional/_intensity.py
```diff
@@ -48,3 +50,64 @@ def _calculate_courtyards(group):
...
+    required_cols = ["node_start", "node_end"]
+    for col in required_cols:
+        if col not in edges.columns:
+            raise ValueError(f"Column {col} is needed in the edges GeoDataframe.")
+
+    if weighted and ("degree" not in nodes.columns):
+        raise ValueError("Column degree is needed in nodes GeoDataframe.")
+
+    def _calc_nodedensity(group, edges):
+        """ "Helper function to calculate group values."""
```
jGaboardi, line 100:
```suggestion
        """Helper function to calculate group values."""
```

jGaboardi (approved):
LGTM after the correction of the minor typo.

jGaboardi:
> Potentially, the function can be sped up further if we change the function API and pass the networkx.MultiGraph object.
Then, we can do:
>
> `sum(l for (,,l) in nx.subgraph(neighbours).edges.data('mm_len'))`
>
> where neighbours comes from the new libpysal.Graph.

Was there any more thought to this?

martinfleis:
> Was there any more thought to this?

Yes. It was faster but the API got super weird so we scraped it.

PR #578, opened 2024-05-12
linting & formatting for `benchmarks/*` and `docs/*`

### LINKED ISSUE #577: more complete linting & formatting - docs & benchmarks
@martinfleis How do yo think about adding linting & formatting for the `docs/*` and `benchmarks/*` directories? I can take care of that.

jGaboardi:
* This PR resolves #577 
* ***big diff, but only linting & formatting***


.pre-commit-config.yaml
```diff
@@ -1,7 +1,7 @@
-files: 'momepy\/'
+files: 'docs\/|benchmarks\/|momepy\/'
```
martinfleis, line 1:
```suggestion
```

We can probably remove this altogether at this point, no?

martinfleis (commented):
Can we set different line length for docs? Because using 88, some cells won't fit, like the first one after [this section](https://momepy--578.org.readthedocs.build/en/578/examples/clustering.html#Streets). I would like to avoid this in the documentation.  It seems that the limit there is 80 characters.

jGaboardi:
> [this section](https://momepy--578.org.readthedocs.build/en/578/examples/clustering.html#Streets)

Yes, seems like we can [accomplish this](https://docs.astral.sh/ruff/configuration/#config-file-discovery) with a subset `ruff.toml` in the `docs/` directory.

.pre-commit-config.yaml
jGaboardi: Sure. I'll give it a shot.

PR #580, opened 2024-05-16
Functional blocks

u3ks:
functional implementation of blocks, the code is pretty much the same as the OO version


momepy/functional/_elements.py
```diff
@@ -343,3 +345,89 @@ def get_nearest_street(
 
     ids[blg_idx] = streets.index[str_idx]
     return ids
+
+
+def generate_blocks(
+    tessellation: GeoDataFrame,
+    edges: GeoDataFrame,
+    buildings: GeoDataFrame,
```
martinfleis, line 353: We don't need this as it can be retrieved from tessellation which should have an index matching with buildings. a simple assignment would do the trick there.

momepy/functional/_elements.py
```diff
@@ -343,3 +345,89 @@ def get_nearest_street(
...
+    id_name: str = "bID",
```
martinfleis, line 354: no need for that, rely on index

momepy/functional/_elements.py
```diff
@@ -343,3 +345,89 @@ def get_nearest_street(
...
+):
```
martinfleis, line 355: missing output type

martinfleis (changes requested):

momepy/functional/_elements.py
```diff
@@ -354,6 +356,86 @@ def get_nearest_street(
...
+        predicate="within",
+    )
+    cells_copy = tessellation[[tessellation.geometry.name]].merge(
+        centroids_temp_id[["component"]], right_index=True, left_index=True, how="left"
+    )
+    blocks = cells_copy.dissolve(by="component").explode(ignore_index=True)
+
+    # assign block ids to buildings and tessellations
+    centroids_w_bl_id2 = gpd.sjoin(buildings_c, blocks, how="left", predicate="within")
+    # return blocks, centroids_w_bl_id2
```
martinfleis, line 428:
```suggestion

```

martinfleis (commented):
We should also make a note that this assumes morphological tessellation and 1:1 relationship between buildings and cells. With enclosed tessellation, those cells that do not have buildings would cause a trouble here. No need to change anything, just adding a note to the docstring would do.

martinfleis (approved):

PR #585, opened 2024-05-20
Functional count

u3ks:
Functional implmentation of mm.Count

this function has the same logic as area_ratio, but the weight is inferred from the geometry type.
Potentially, it could be left to the user to implement it with a call to describe_reached (or two if the result should be weighted). (similarly to arearatio)


momepy/functional/_intensity.py
```diff
@@ -111,3 +111,65 @@ def _calc_nodedensity(group, edges):
         summation_values = pd.Series(np.ones(nodes.shape[0]), index=nodes.index)
 
     return graph.apply(summation_values, _calc_nodedensity, edges=edges)
+
+
+def count(
+    left: GeoDataFrame,
+    right: GeoDataFrame,
+    right_group_key: Series | np.ndarray,
```
martinfleis, line 119:
```suggestion
    aggregation_key: Series | np.ndarray,
```
as per  #583

momepy/functional/_intensity.py
```diff
@@ -111,3 +111,65 @@ def _calc_nodedensity(group, edges):
         summation_values = pd.Series(np.ones(nodes.shape[0]), index=nodes.index)
 
     return graph.apply(summation_values, _calc_nodedensity, edges=edges)
+
+
+def count(
+    left: GeoDataFrame,
+    right: GeoDataFrame,
```
martinfleis, line 118: Do we actually need `right` at all? Doesn't the grouping have all the info we need?

martinfleis (changes requested):

PR #587, opened 2024-05-20
Functional dimension

u3ks:
functional dimension implementation, mostly the same code with for loops replaced by graph.apply


momepy/functional/_dimension.py
```diff
@@ -157,3 +158,72 @@ def perimeter_wall(
     ].values
 
     return results
+
+
+def weighted_character(
+    values: np.ndarray | Series, areas: np.ndarray | Series, graph: Graph
```
martinfleis, line 164:
```suggestion
    y: NDArray[np.float_] | Series, area: NDArray[np.float_] | Series, graph: Graph
```
proper typing, no plural in args.

momepy/functional/_dimension.py
```diff
@@ -157,3 +158,72 @@ def perimeter_wall(
...
+        A Series containing the resulting values.
+
+    Examples
+    --------
+    >>> res = mm.weighted_character(buildings_df['height'],
+    ...                     buildings_df.geometry.area, graph)
+    """
+    data = np.vstack((values, areas)).T
+    stats = graph.apply(
+        data, lambda x: np.sum(x.iloc[:, 0] * x.iloc[:, 1]) / np.sum(x.iloc[:, 1])
```
martinfleis, line 197: If both are arrays, there's no `iloc`. Either cast it to df or to array.

momepy/functional/_dimension.py
```diff
@@ -157,3 +158,72 @@ def perimeter_wall(
...
+    )
+
+    result = Series(np.nan, index=graph.unique_ids)
+    result.loc[stats.index.values] = stats.values
```
martinfleis, line 201: This is done why?

momepy/functional/_dimension.py
```diff
@@ -157,3 +158,72 @@ def perimeter_wall(
...
+    return result
+
+
+def covered_area(areas: np.ndarray | Series, graph: Graph) -> Series:
```
martinfleis, line 205:
```suggestion
def covered_area(area: np.ndarray | Series, graph: Graph) -> Series:
```

martinfleis (changes requested):
One thing that applies to all of the PRs. If the type hinting suggest some types, we should have tests using those. You are often using code that works ony for one of those types.

momepy/functional/_dimension.py
```diff
@@ -157,3 +157,40 @@ def perimeter_wall(
     ].values
 
     return results
+
+
+def weighted_character(values: Series, areas: Series, graph: Graph) -> Series:
```
martinfleis, line 162:
```suggestion
def weighted_character(y: Series, areas: Series, graph: Graph) -> Series:
```

Custom variable shall always be `y` and shall support both numpy array and Series.

martinfleis (changes requested):

martinfleis (approved):

PR #589, opened 2024-05-20
Functional percentiles

u3ks:
I seperated the percentiles function into weighted and unweighted, since the logic is different, otherwise the code is similar to the original implementation.


momepy/functional/_diversity.py
```diff
@@ -302,3 +329,44 @@ def describe_reached(
...
+    ----------
+    y : Series
+        A Series containing the values to be analysed.
+    graph : libpysal.graph.Graph
+        A spatial weights matrix for the data.
+    q : array-like (default [25, 50, 75])
+        The percentiles to return.
+
+    Returns
+    --------
```
martinfleis, line 354:
```suggestion
    -------
```

momepy/functional/_diversity.py
```diff
@@ -302,3 +329,44 @@ def describe_reached(
...
+
+def percentile(
+    y: Series,
+    graph: Graph,
+    q: tuple | list = [25, 50, 75],
+):
+    """Calculates linearly weighted percentiles of ``y`` values using
+    the neighbourhoods and weights defined in ``graph``.
+
+    The specific interpolation method implemented is "hazen".
```
martinfleis, line 342: What does that mean?

momepy/functional/tests/test_diversity.py
```diff
@@ -279,6 +302,80 @@ def test_na_results(self):
 
         assert_frame_equal(pandas_agg_vals, numba_agg_vals)
 
+    @pytest.mark.skipif(not GPD_013, reason="get_coordinates() not available")
```
martinfleis, line 305: Where do you use `get_coordinates`?

momepy/functional/_diversity.py
```diff
@@ -302,3 +329,44 @@ def describe_reached(
...
+
+    weights = graph._adjacency.values
+    vals = y.loc[graph._adjacency.index.get_level_values(1)]
+    vals = np.vstack((weights, vals)).T
+    vals = DataFrame(vals, columns=["weights", "values"])
+    grouper = vals.groupby(graph._adjacency.index.get_level_values(0))
+    q = tuple(q)
+    stats = grouper.apply(lambda x: _interpolate(x.values, q))
+    result = DataFrame(np.stack(stats), columns=q, index=stats.index)
+    return result
```
martinfleis, line 372: We need to treat isolates as we do in `describe`.

momepy/functional/_diversity.py
u3ks: its one of the interpolation method in np.interpolate, and the one implemented in momepy.Percentile(weights='linear_weighted').
Basically we divide the weights/xs by their sum .

momepy/functional/tests/test_diversity.py
u3ks: changed

martinfleis (approved):

PR #598, opened 2024-06-13
DOC: note on a precision issue in enclosed_tessellation

martinfleis:
We have noticed a nasty issue with the contiguity of enclosed tessellation caused by floating point imprecision within GEOS. Spent some time trying to fix it but I don't think it is possible with the current version of GEOS / shapely. For our use case, there is a solution, so making a note in the docstring.

Repr of the issue:

```py
import geopandas as gpd
import momepy

streets = gpd.read_file(momepy.datasets.get_path("bubenec"), layer="streets")
buildings = gpd.read_file(momepy.datasets.get_path("bubenec"), layer="buildings")

enclosures = momepy.enclosures(streets, limit=momepy.buffered_limit(buildings, buffer=50))

tess = momepy.enclosed_tessellation(buildings, enclosures)

tess.union_all()
```
Screenshot 2024-06-13 at 22 00 25

The lines shown in the interior are actual gaps in contiguity when building the graph:

```py
from libpysal.graph import Graph

m = tess_new.explore()
g = Graph.build_contiguity(tess_new, rook=False, strict=True)
g.explore(tess_new, m=m)
m
```
Screenshot 2024-06-13 at 22 02 06

It's been in the enclosed tessellation since the first implementation 4 years ago.

### DIFF
--- a/momepy/functional/_elements.py
+++ b/momepy/functional/_elements.py
@@ -159,7 +159,7 @@ def enclosed_tessellation(
     coverage. To build a contiguity graph, use fuzzy contiguity builder with a small
     buffer, e.g.::
 
-        from libpysal imoprt graph
+        from libpysal import graph
 
         graph.Graph.build_fuzzy_contiguity(tessellation, buffer=1e-8)
 
--- a/momepy/functional/_elements.py
+++ b/momepy/functional/_elements.py
@@ -43,15 +43,15 @@ def morphological_tessellation(
     and there are three particular patterns causing issues:
 
     1. Features will collapse into empty polygon - these
-        do not have tessellation cell in the end.
+       do not have tessellation cell in the end.
     2. Features will split into MultiPolygons - in some cases,
-        features with narrow links between parts split into two
-        during 'shrinking'. In most cases that is not an issue
-        and the resulting tessellation is correct anyway, but
-        sometimes this results in a cell being a MultiPolygon,
-        which is not correct.
+       features with narrow links between parts split into two
+       during 'shrinking'. In most cases that is not an issue
+       and the resulting tessellation is correct anyway, but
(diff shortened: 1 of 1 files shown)


martinfleis:
Just for a reference, this is how it should ideally look like.

Screenshot 2024-06-13 at 22 13 34

jGaboardi:
Is this something we are going to report upstream to `shapely`?

momepy/functional/_elements.py
```diff
@@ -152,6 +152,17 @@ def enclosed_tessellation(
         The number of jobs to run in parallel. -1 means using all available cores.
         By default -1
 
+    Warnings
+    --------
+    Due to the floating point precision issues in clipping the tessellation cells to the
```
jGaboardi, line 157: Should we include anything in here about the specific `shapely`/`GEOS` versions that lead to this issue?

martinfleis: All of them from my testing.

martinfleis:
> Is this something we are going to report upstream to shapely?

It is on my to-do list but I've been a bit demoralised by this finding so haven't managed to find a strength to come up with a minimal repr.

momepy/functional/_elements.py
u3ks: I tried some old gds_py containers with various versions of shapely and the problem was still there

jGaboardi (approved):

momepy/functional/_elements.py
martinfleis: Same in the tessellation generated in 2020.

jGaboardi: 😬 oh no

martinfleis:
It's apparently expected... https://github.com/libgeos/geos/issues/1109

PR #599, opened 2024-06-14
DOC: User guide refresh

martinfleis:
Sorry, this is a bit large PR. I have updated the notebooks in the user guide to use the new API.

Alongside I've noticed few things related to MultiIndex:

- some shape stuff did not work
- tessellation seems to break `MultiIndex` but I need to have a proper look
- there was a minor issue in `verify_tessellation`
- `get_node_id` needs updating to rely in indices like everything else. probably as `get_nearest_node`


martinfleis:
@jGaboardi I think that this is mergeable now. there are some follow-ups but this is already large enough

docs/user_guide/elements/links.ipynb
jGaboardi, line 1: Cell 14 fails here. Is this expected?

martinfleis: Yes, I need to re-do the function and will update that later.

jGaboardi (approved):

PR #600, opened 2024-06-14
ENH: add get_nearest_node

martinfleis:
Index-oriented and vectorized alternative of `get_node_id`. Also fixes the user guide failure.


.github/workflows/test_user_guide.yaml
```diff
@@ -27,7 +27,7 @@ jobs:
       - name: setup micromamba
         uses: mamba-org/setup-micromamba@v1
         with:
-          environment-file: ci/envs/312-latest.yaml
```
martinfleis, line 30: I am using some of the geopandas 1.0 api (union_all). GeoPandas will be released sooner than momepy.

jGaboardi (approved):

PR #601, opened 2024-06-14
DOC: create usable env on RTD, update clustering

martinfleis:
This is an attempt to always execute notebooks when running the documentation build. Until now, they had to be manually executed but I think that the limits we and RTD had before are gone now.

I've also updated clustering example notebook I forgot about before. 

The downside is that building the documentation will now take significantly longer than before but it will always be up to date.


jGaboardi:
> significantly longer

Any idea how long? (I guess we'll find out soon enough)

martinfleis:
14 minutes locally. Given I just learned that there is a limit on RTD

> 15 minutes build time

This is not going to work :)

martinfleis:
I'll try if we can at least avoid those mocking to get a running environment for some small bits we may want to execute.

martinfleis:
Okay, let's call this one done. 

We now have an environment that is able to run stuff if needed but the notebooks are still executed using the "auto" mode, meaning that if they are pushed executed, no execution happens during the build.

The CI issues looks like some temporary network issue. And I haven't touched the code :).

jGaboardi (approved):

PR #602, opened 2024-06-14
API: return morphological tessellation as a GeoDataFrame

martinfleis:
Normally, you will want to use tessellation as a gdf, so easier to return it as such and avoid all those `to_frame` calls. This also makes the output type consistent with `enclosed_tessellation`.


jGaboardi (approved):

jGaboardi:
That failure is one from `osmnx`, not `momepy` itself.

Issue #543, opened 2024-02-01
GHA: switch to autogenerated release notes


martinfleis:

martinfleis:
@jGaboardi this sounds like a James task. Could you have a look into that? You were setting it up elsewhere. We should also remove the rendered changelog from the docs and link to release notes on GH instead.

jGaboardi:
@martinfleis I will try to give this some attention over the weekend.

martinfleis:
I'd go with `0.8.0-alpha{1...N}` as it reflects the best the nature of those releases as per semver. We're not at RC stage yet.

jGaboardi:
#### Steps

1. open pr with new infrastructure - #604 
2. decide on labels/tags for `release.yml`, etc.
3. merge once agreed
4. iterate tag+release over `0.8.0-alpha{1...N}` until satisfied

martinfleis:
+ remove changelog from docs

jGaboardi:
beginning alpha release attempts
* [ ] `v0.8.0-alpha1`

martinfleis:
I have updated tags in the relevant PRs since 0.7.0

jGaboardi:
So we're all good to close this? Are we sure we don't want one final test release to make sure those new tags are picked up properly and looking nice? Probably not necessary, more a peace-of-mind thing.

martinfleis:
> So we're all good to close this?

I guess so?

> Are we sure we don't want one final test release to make sure those new tags are picked up properly and looking nice?

let's YOLO it 😎

PR #604, opened 2024-06-15
gha for release notes - #543

jGaboardi:
gha for release notes - #543

* decide on tags for release
* the [release action for `spopt`](https://github.com/pysal/spopt/blob/main/.github/workflows/release_and_publish.yml) and the [one here in `momepy`](https://github.com/pysal/momepy/blob/main/.github/workflows/release_to_pypi.yml) differ considerably -- will have to see how adding the "create release notes" interacts here
* what else?


.github/release.yml
```diff
@@ -0,0 +1,28 @@
+changelog:
+  exclude:
+    labels:
+      - ignore-for-release
+    authors:
+      - dependabot
```
martinfleis, line 6:
```suggestion
      - dependabot
      - pre-commit-ci
```

.github/workflows/release_to_pypi.yml
```diff
@@ -41,6 +41,16 @@ jobs:
           draft: false
           prerelease: false
 
+      - name: Create Release Notes
```
martinfleis, line 44: I suppose that this will conflict with Create GitHub Release above

martinfleis (commented):
On the categories, something like

- API changes
- Enhancements
- Bug fixes
- Deprecations
- Documentation
- Other

? I am not a fan of having type (e.g. enhancement) and content (e.g. preprocessing) mixed together. I also don't think that refactoring needs to be included as it should not affect user in any way.

jGaboardi:
> On the categories, something like
> 
> * API changes
> * Enhancements
> * Bug fixes
> * Deprecations
> * Documentation
> * Other
> 
> ? I am not a fan of having type (e.g. enhancement) and content (e.g. preprocessing) mixed together. I also don't think that refactoring needs to be included as it should not affect user in any way.

So maybe like:

* categories
  * title --  **API changes**
    * label -- `refactor`?
  * title --  **Enhancements**
    * label -- `enhancements`
  * title -- **Bug fixes**
    * label -- `bug`
  * title -- **Deprecations**
    * label -- ?
  * title -- **Documentation**
    * label -- `documentation` 
  * title -- **Other**
    * label -- `*` 

The thing is that we don't have labels for all and will have go back through all Issues & PRs to add those new labels once they are created.

.github/workflows/release_to_pypi.yml
jGaboardi: > I suppose that this will conflict with Create GitHub Release above

Yeah, more than likely. We will have to work that out iteratively seems like.

.github/release.yml
```diff
@@ -0,0 +1,28 @@
...
+      - pre-commit-ci
+  categories:
+    - title:  API changes
+      labels: refactor, ...
```
jGaboardi, line 10: see [comment](https://github.com/pysal/momepy/pull/604#issuecomment-2169962751)

.github/release.yml
```diff
@@ -0,0 +1,28 @@
...
+    - title: Enhancements
+      labels:
+        - enhancement
+    - title: Bug Fixes
+      labels:
+        - bug
+    - title: Deprecations
+      labels:
+        - ...
```
jGaboardi, line 19: see [comment](https://github.com/pysal/momepy/pull/604#issuecomment-2169962751)

martinfleis:
> The thing is that we don't have labels for all and will have go back through all Issues & PRs to add those new labels once they are created

I am happy to do that.


* categories
  * title --  **API changes**
    * label -- `refactor`? - refactor is not API change. API change is stuff like change of a default arg value. We don't have a label for that and we haven't done any API changes. All stuff regarding the API model is considered enhancement as those are new functions that live alongside the old.
  * title --  **Enhancements**
    * label -- `enhancements`
  * title -- **Bug fixes**
    * label -- `bug`
  * title -- **Deprecations**
    * label -- ? - we need a label `deprecation`
  * title -- **Documentation**
    * label -- `documentation` 
  * title -- **Other**
    * label -- `*`

jGaboardi:
> refactor is not API change. API change is stuff like change of a default arg value. We don't have a label for that and we haven't done any API changes. All stuff regarding the API model is considered enhancement as those are new functions that live alongside the old.

Roger that. OK, I'll add `api` and `deprecation` labels within `release.yml` then go ahead and create those labels in the repo.

.github/workflows/release_to_pypi.yml
martinfleis: We need one or the other. Also note that removing the one above will break the last one which uses `steps.create_release.outputs.upload_url`. In any case, `actions/create-release` is archived so we shall find a new way of doing things.

jGaboardi: I actually think things are more simple than I thought. We can simply prune down the current `momepy` action to get our desired result. That final step []()

We can see in the [`v0.7.0` release](https://github.com/pysal/momepy/releases/tag/v0.7.0) there are 3 assets there. 
* `momepy-0.7.0.tar.gz`
* `Source code (zip)`
* `Source code (tar.gz)`

However, `momepy-0.7.0.tar.gz` and `Source code (tar.gz)` are the exact same thing, just different names - that's what [*Upload Release Asset (sdist) to GitHub*](https://github.com/pysal/momepy/blob/e6b0abf4c195d5c7908355e624cce96fd83bde02/.github/workflows/release_to_pypi.yml#L50) is doing (and it's not needed).

I'm going to make some adjustments and push up (what I hope is) the fix.

jGaboardi:
@martinfleis this is ready for review. once merged I will see about a tag + release of `v0.8.0-alpha1`

jGaboardi:
Same `osmnx` error causing failures -- https://github.com/pysal/momepy/actions/runs/9530670355/job/26270797522?pr=604#step:5:544

martinfleis (approved):

PR #608, opened 2024-06-17
BUG: fix describe_ function when count is not present

u3ks:
a fix for a small issue with the mm.describe_* functions, where they expect a count to always be calculated


martinfleis (approved):

martinfleis:
Is this something that also needs to be included in Graph.describe?

u3ks:
No, Graph.describe doesnt modify the output of pandas.agg, here we do to make it compatible with the old functionality

PR #609, opened 2024-06-17
Faster node density

u3ks:
I moved the node density functionality to the graph module and use networkx to compute it. 
The current implementation scales horribly and on larger graphs takes hours, since all edges have to be filtered multiple times with np.in1d.
Getting the exact timing is a bit messy since subgraphs and higher order pysal have to be precomuted and affect the time.
The new implementation takes around <20ish mintues on a graph with ~500k nodes and 2mil edges, together with the other mm.subgraph computations.
For smaller datasets the current implementation  + calculating a higher order graph ~ the new networkx implementation.


momepy/graph.py
```diff
@@ -988,6 +990,72 @@ def straightness_centrality(
     return netx
 
 
+def node_density(
+    graph: nx.Graph,
+    radius: int,
+    length: str = "mm_len",
+    distance: str | None = None,
+    verbose: bool = True,
+):
```
martinfleis, line 999:
```suggestion
) -> nx.Graph:
```

momepy/graph.py
```diff
@@ -988,6 +990,72 @@ def straightness_centrality(
...
+        Include all neighbors of distance <= radius from ``n``.
+    length : str, default `mm_len`
+        The name of the attribute of segment length (geographical).
+    distance : str, optional
+        Use specified edge data key as distance.
+        For example, setting ``distance=’weight’`` will use the edge ``weight`` to
+        measure the distance from the node ``n`` during ``ego_graph`` generation.
+    verbose : bool (default True)
+        If ``True``, shows progress bars in loops and indication of steps.
+    Returns
```
martinfleis, line 1026:
```suggestion

    Returns
```

momepy/graph.py
```diff
@@ -988,6 +990,72 @@ def straightness_centrality(
...
+    -------
+    netx : Graph
+        A networkx.Graph object.
+
+    Examples
+    --------
+    >>> nodes['density'] = mm.node_density(nodes, edges, graph)
```
martinfleis, line 1033: this no longer applies

martinfleis (commented):
Cool!

momepy/graph.py
```diff
@@ -1031,7 +1031,7 @@ def node_density(
 
     Examples
     --------
-    >>> nodes['density'] = mm.node_density(nodes, edges, graph)
+    >>> network_graph = mm.straightness_centrality(network_graph, radius=5)
```
martinfleis, line 1034: This also doesn't apply :D

u3ks: i changed the radius, but not the name....

martinfleis (approved):

martinfleis:
One more thing, can you update API docs? Removing the old one and adding the new to the proper section.

PR #613, opened 2024-06-18
DOC: ruff user guide

martinfleis:
@jGaboardi any idea why it was not picked up by pre-commit?


jGaboardi:
Not 100% sure, but let's try to explicitly add the path in `.pre-commit-config.yaml`, like [here in `spopt`](https://github.com/pysal/spopt/blob/fa782a5cb0ead740f3776c2f305522014515f5eb/.pre-commit-config.yaml#L1). So ours would look like:

```yaml
files: "momepy\/|docs\/"
```

PR #614, opened 2024-06-18
update precommit to ruff docs dir

jGaboardi:
update precommit to ruff docs dir --  xref #613


martinfleis (approved):
Ah, that is what was missing :).

PR #617, opened 2024-06-18
API: allow silencing of FutureWarnings from legacy API

martinfleis:
This introduces an environment variable `ALLOW_LEGACY_MOMEPY` which, when set to True, silences all the deprecation warnings about the legacy API.

I am open to a discussion about the name of the variable but I'd say anything without a possibility of conflicts is good enough.

### DIFF
--- a/docs/legacy_api.rst
+++ b/docs/legacy_api.rst
@@ -20,7 +20,7 @@ elements
 --------
 
 .. autosummary::
-   :toctree: generated/a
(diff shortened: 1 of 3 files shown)


jGaboardi (approved):
I think `ALLOW_LEGACY_MOMEPY` is good enough. Maybe to be more concise we could do `USE_LEGACY_MOMEPY`, but the difference is trivial.

PR #619, opened 2024-06-19
API: deprecate legacy API in favour of Graph-based functional implementation

### LINKED ISSUE #612: API: deprecate legacy API
We need to properly use the decorators to depreciate all of the Legacy API.

On top, we should probably add either env. variable or some global setting allowing folks to silence all of those warnings, otherwise a typical workflow would be very noisy.

martinfleis:
Closes #612 

Adds decorators and custom warnings when decorator does not fit to remaining classes.

The only missing is `NeighboringStreetOrientationDeviation`. @u3ks what are we doing about that one?

@u3ks also, can you review this if I'm pointing to correct things?

### DIFF
--- a/momepy/elements.py
+++ b/momepy/elements.py
@@ -2,6 +2,7 @@
 
 # elements.py
 # generating derived elements (street edge, block)
+import os
 import warnings
 
 import geopandas as gpd
@@ -14,6 +15,8 @@ from scipy.spatial import Voronoi
 from shapely.geometry.base import BaseGeometry
 from shapely.ops import polygonize
 from tqdm.auto import tqdm
+
+from .utils import deprecated
 
 __all__ = [
     "Tessellation",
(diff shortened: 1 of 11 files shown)


momepy/tests/test_dimension.py
```diff
@@ -18,67 +18,30 @@ def setup_method(self):
         self.df_buildings["height"] = np.linspace(10.0, 30.0, 144)
 
     def test_Area(self):
-        with pytest.warns(
```
martinfleis, line 21: Given we test the decorators and can silence these warnings, I am removing explicit checks from individual classes.

momepy/elements.py
```diff
@@ -170,6 +173,22 @@ def __init__(
         use_dask=True,
         n_chunks=None,
     ):
+        if os.getenv("ALLOW_LEGACY_MOMEPY", "False").lower() not in (
```
jGaboardi, line 176: Although this message varies across differing classes, perhaps one message with formatted bits stored somewhere might be a good idea? Something like:

```
(
    "Class based API like `momepy.{}` is deprecated. "
    "Replace it with `momepy.{}` or "
    "`momepy.{}` to use functional API instead "
    "or pin momepy version <1.0. Class-based API will be removed in "
    "1.0. "
).format(....)
```

jGaboardi: I'm in the fence now that it's only used in 3 places. Probably OK the way it is unless there are more places we'll need to warn.

jGaboardi (approved):

momepy/elements.py
martinfleis: All three are unique, there's no point in making a template.

momepy/intensity.py
```diff
@@ -135,6 +136,7 @@ def __init__(
         self.series = objects_merged["lf_area"] / objects_merged[left_areas]
 
 
+@deprecated("describe_agg")
```
u3ks, line 139: Shouldnt this be removed instead?

u3ks:
NeighboringStreetOrientationDeviation is missing.
I think the same functionality can be reimplemented using graph and then a call to Graph.apply.
But I dont think we can leave it to the user, since we got to define a custom apply function to take absolute difference, then mean.

martinfleis:
> NeighboringStreetOrientationDeviation is missing.

Can you look into that one?

PR #622, opened 2024-06-20
ENH: either support MultiIndex or raise an error when one is given

### LINKED ISSUE #606: BUG: verify handling of MultiIndex
It seems that when your GeoDataFrame is using MultiIindex, it does affect some parts of `momepy` and will likely break when using Graph. We should test and verify that it behaves as we want, both here and in libpysal. In some cases, raising may be the best option but if we can support it, it is better. I don't think that there's scope in supporting MultiIndex in Graph so it may need to be checked only in functions that do not use graphs.

u3ks:
Closes #606


momepy/functional/_elements.py
```diff
@@ -103,6 +103,10 @@ def morphological_tessellation(
     4  POLYGON ((1603084.231 6464104.386, 1603083.773...
 
     """
+
+    if isinstance(geometry.index, MultiIndex):
+        raise ValueError("MultiIndex is not supported for tessellation.")
```
martinfleis, line 108:
```suggestion
        raise ValueError("MultiIndex is not supported in momepy.morphological_tessellation.")
```

momepy/functional/_elements.py
```diff
@@ -218,6 +222,9 @@ def enclosed_tessellation(
     126  POLYGON ((1603528.593 6464221.033, 1603527.796...                0
     """
 
+    if isinstance(geometry.index, MultiIndex):
+        raise ValueError("MultiIndex is not supported for tessellation.")
```
martinfleis, line 226:
```suggestion
        raise ValueError("MultiIndex is not supported in `momepy.enclosed_tessellation`.")
```

momepy/functional/_elements.py
```diff
@@ -360,6 +367,12 @@ def verify_tessellation(tessellation, geometry):
 
     >>> excluded, multipolygons = momepy.verify_tessellation(tessellation, buildings)
     """
+
+    if isinstance(geometry.index, MultiIndex) or isinstance(
+        tessellation.index, MultiIndex
+    ):
+        raise ValueError("MultiIndex is not supported for tessellation.")
```
martinfleis, line 374:
```suggestion
        raise ValueError("MultiIndex is not supported in `momepy.verify_tessellation`.")
```

momepy/functional/_elements.py
```diff
@@ -397,7 +410,7 @@ def get_nearest_street(
     buildings: GeoSeries | GeoDataFrame,
     streets: GeoSeries | GeoDataFrame,
     max_distance: float | None = None,
-) -> np.ndarray:
+) -> Series:
```
martinfleis, line 413: This will need an update of example

momepy/functional/_elements.py
```diff
@@ -522,6 +534,17 @@ def get_nearest_node(
...
     """
+
+    if (
+        isinstance(buildings.index, MultiIndex)
+        or isinstance(nearest_edge.index, MultiIndex)
+        or isinstance(nodes.index, MultiIndex)
+        or isinstance(edges.index, MultiIndex)
+    ):
+        raise ValueError(
+            "MultiIndex is not supported for calculating the nearest node."
```
martinfleis, line 545:
```suggestion
            "MultiIndex is not supported in `momepy.get_nearest_node`."
```

momepy/functional/_elements.py
```diff
@@ -606,6 +629,12 @@ def generate_blocks(
     >>> tessellation["block_id"] = tessellation_id
     """
 
+    if (
+        isinstance(buildings.index, MultiIndex)
+        or isinstance(tessellation.index, MultiIndex)
+        or isinstance(edges.index, MultiIndex)
+    ):
+        raise ValueError("MultiIndex is not supported for generating blocks.")
```
martinfleis, line 637:
```suggestion
        raise ValueError("MultiIndex is not supported in `momepy.generate_blocks`.")
```

martinfleis (approved):

PR #630, opened 2024-06-21
[COINS] Fix angle between lines

fnattino:
Giving it a try to a version that simplifies the calculation of angles between the line segments.


martinfleis:
@anastassiavybornova could you review this one please?

anastassiavybornova:
@martinfleis yes will do

momepy/coins.py
```diff
@@ -378,118 +378,37 @@ def _list_to_pairs(in_list):
...
-        return -_compute_angle(point1, point2)
-
-    # if the latitudes are same, the line is horizontal
-    elif point2[1] == point1[1]:
-        return 0
-
-    # if the longitudes are same, the line is vertical
-    elif point2[0] == point1[0]:
-        return 90
+    """Calculates the orientation of a line segment."""
```
fnattino, line 382: This function behaves slightly differently from the earlier version, in that it accounts for the vector direction as well (maybe I should actually rename it to `_compute_direction`?). Thus, the return values now span the range (-180;180], instead of instead of (-90;90] of the previous version of this function. 

While I have implemented this change to reuse this function in `_compute_angle`, I have noticed that this function is also called [here](https://github.com/pysal/momepy/blob/c3c5d5ee704a2fa5d3046e2b4be14c0832d93be9/momepy/coins.py#L129) and the vector orientation is added to the dict of information for the line segments. As far as I can tell, however, this spefic info on the orientation is not used anywhere else - so I guess it should be fine if the direction rather the orientation is reported in there?

momepy/coins.py
```diff
@@ -378,118 +378,37 @@ def _list_to_pairs(in_list):
...
-        elif l2orien < 0:
-            return 180 - abs(l2orien)
-        else:
-            return 180 - (
-                abs(_compute_orientation(line1)) + abs(_compute_orientation(line2))
-            )
-
-    elif l1orien == l2orien:
-        return 180
+def _compute_angle(a, b, c):
```
fnattino, line 390: I have noticed there was also a function to compute the angle included between three points here:

https://github.com/pysal/momepy/blob/c3c5d5ee704a2fa5d3046e2b4be14c0832d93be9/momepy/utils.py#L103-L110

However, to get the angle included between the points ("a", "b", "c") with "b" being the vertex I had to switch the order of "a" and "b" - or am I overlooking something?

momepy/coins.py
fnattino: Or maybe, even better, I could rename this `_compute_direction` and keep using it in `_compute_angle`, then add `_compute_orientation` as the following, to be used in `_split_lines()`:

```python
def _compute_orientation(line):
    point1 = line[0]
    point2 = line[1]
    dy = point1[1] - point2[1]
    if dy == 0.:
        return 0.
    dx = point1[0] - point2[0]
    if dx == 0.:
        return 90.
    return math.degrees(math.atan(dy/dx))
```

anastassiavybornova: @fnattino thanks for this! i agree with you in that the orientation is not used anywhere so not really necessary for the angle computation between two linestrings. in fact, since by definition the linestrings that we're computing angles ("interior angle") between will always (by definition) start in the same point, we don't need neither orientation *nor* direction -  just the good old angle-between-two-vectors formula which will give an angle 0<=angle<=180. in my understanding, the original COINS implementation was using orientation because the interior angle was being computed in reference to the horizontal plane. but if instead of referencing the horizontal plane, we just make use of the fact that we're essentially dealing with 2 vectors from the same origin, the computation becomes way easier (no use cases have to be distinguished as to which quadrant we're operating in)

momepy/coins.py
anastassiavybornova: (see also my comment above) if we're using the vector logic, we don't need to compute angles between points (again, this is an artifact of having the horizontal plane as reference) and can go directly to computing angles between vectors. and since by definition the 2 vectors start in the same point, we can define their direction as both pointing away from the origin. so @fnattino exactly, agree that the order is relevant, *however* we luckily don't need to consider this if we use vector logic for angle computation

PR #632, opened 2024-06-27
BUG: refactor computation of angles in COINS

### LINKED ISSUE #629: [COINS] angles between segments in the same quadrant of the Cartesian plane
### Describe the problem

Hi there, thanks a lot for this fantastic library!

I am playing with momepy's  COINS implementation and I have identified some strange behaviour when creating strokes over edges that connect with some sharp angles.  I have narrowed down the issue to what I believe being wrong angles being calculated for segments lying in the same quadrant of the Cartesian plane (see examples below). 

If I understand  `_angle_between_two_lines` correctly, one checks whether two segments lie on the same side of the horizontal plane (e.g. [here](https://github.com/pysal/momepy/blob/c3c5d5ee704a2fa5d3046e2b4be14c0832d93be9/momepy/coins.py#L463-L464)). If this is found to be the case, the result seems to assume that the two segments lie on opposite sides of the vertical plane (`angle = 180 - (abs(l1orien) + abs(l2orien))`). The case with the segments lying on the same side of the vertical plane (`angle = abs(l1orien - l2orien)`) seems not to be covered. Or am I missing something here? 

Happy to contribute with a PR if you confirm this is unwanted behaviour!

### Steps to reproduce

For the following pair of segments, I get these angles:

```python
import momepy.coins

l1 = [[0., 0.], [0.1, 0.2]]
l2 = [[0., 0.], [0.2, 0.1]]
angle_l12 = momepy.coins._angle_between_two_lines(l1, l2)
print(angle_l12)
# 90.0

l3 = [[-0.4, -0.1], [0., 0.]]
l4 = [[-0.6, -0.2], [0., 0.]]
angle_l34 = momepy.coins._angle_between_two_lines(l3, l4)
print(angle_l34)
# 147.529

```

while I would have expect instead something like:

```python
import numpy as np

def normalize(x):
    return np.array(x) / np.linalg.norm(x)

def angle_between(l1, l2):
    v1 = np.array(l1)
    v1 -= v1[0]
    v2 = np.array(l2)
    v2 -= v2[0]
    return np.degrees(np.arccos(np.dot(
        normalize(v1[1]), 
        normalize(v2[1])
    )))

angle_l12 = angle_between(l1, l2)
print(angle_l12)
# 36.86989764584403

angle_l34 = angle_between(l3, l4)
print(angle_l34)
# 4.398705354995591
````
 
Visualization of the segments:

image

image

### Versions of your packages

momepy=0.6.0

### Your operating system

macOS - x86-64

### Additional context

_No response_

### PR DESCRIPTION
**Pull request by anastassiavybornova**
Based on #629 (solved) and #630 (superseded)
Refactor computation of interior angle between 2 lines.

Previously, the interior angle between 2 lines was  computed by:
1. computing each line's angle formed with the *horizontal plane*
2. computing each line's orientation relative to the *horizontal plane*
3. applying trigonometric formulas while distinguishing different use cases (of orientation combinations)

this is now simplified into making use of the fact that by definition, the two lines we're computing interior angles between will always start in the same point, hence we can reduce the entire problem to the computation of an angle between 2 vectors in 2dim space from the same origin, hence referencing the horizontal plane // orientation // direction becomes obsolete. this pr supersedes #630 in that orientation is dropped completely from the code.

just for the records, the bug mentioned in #629 resulted from the fact that the use case of sharp angles (<90deg) was not considered in the use case distinction.

thanks to @fnattino for the input!!

Fixes #629 and closes #630

### CODE REVIEW

File: momepy/coins.py [hunk 1]
```diff
@@ -377,119 +381,59 @@ def _list_to_pairs(in_list):
...
-        else:
-            return 180 - (
-                abs(_compute_orientation(line1)) + abs(_compute_orientation(line2))
-            )
+    # make sure lines are not identical
+    if Counter(points.values()) == {2: 2}:
+        raise ValueError(
+            "ValueError: lines are identical. Please revise input gdf\
+            to ensure no lines are identical or overlapping."
+        )
```
martinfleis, line 416:
```suggestion
        raise ValueError(
            "Lines are identical. Please revise input data "
            "to ensure no lines are identical or overlapping. "
            "You can check for duplicates using `gdf.geometry.duplicated()`."
        )
```

momepy/coins.py
```diff
@@ -377,119 +381,59 @@ def _list_to_pairs(in_list):
...
 
-    elif l1orien == l2orien:
-        return 180
+    # make sure lines do touch
+    if Counter(points.values()) == {1: 4}:
+        raise ValueError("ValueError: lines do not touch.")
```
martinfleis, line 420:
```suggestion
        raise ValueError("Lines do not touch.")
```

No need to write "ValueError" in the message. That would result in "**ValueError:** ValueError: message"

momepy/coins.py
```diff
@@ -377,119 +381,59 @@ def _list_to_pairs(in_list):
...
 
+    assert Counter(points.values()) == {1: 2, 2: 1}
```
martinfleis, line 422: Can we remove this assertion?

anastassiavybornova: We're accounting for the unwanted cases before that, so yes we can remove it

momepy/preprocessing.py
anastassiavybornova, line 1: changes come from running black on all files

momepy/shape.py
anastassiavybornova, line 1: changes come from running black on all files

momepy/coins.py
```diff
@@ -378,118 +382,51 @@ def _list_to_pairs(in_list):
...
-            return 180 - abs(l1orien)
-        elif l2orien < 0:
-            return 180 - abs(l2orien)
-        else:
-            return 180 - (
-                abs(_compute_orientation(line1)) + abs(_compute_orientation(line2))
-            )
+    # assertion: we expect exactly 2 of the 4 points to be identical
+    # (lines touch in this point)
+    points = Counter([a, b, c, d])
```
anastassiavybornova, line 401: one of (many!!) options to figure out which two of the four points in the vectors (a,b) and (c,d) are identical

momepy/coins.py
```diff
@@ -126,7 +130,7 @@ def _split_lines(self):
                 out_line.append(
                     [
                         part,
-                        _compute_orientation(part),
+                        [],
```
anastassiavybornova, line 133: since the COINS algorithm operates with position-based (index-based) attributes of lines, the very first attribute being "orientation": this is now obsolete, so this list will always be empty (we're not computing orientation anymore), but keeping it there nevertheless for the moment not to mess up with the index-based accessing of attributes in the rest of the algorithm. (would probably require major refactoring if we wanted to get rid of this issue, for now i guess it's best just not to touch it)

momepy/coins.py
```diff
@@ -37,10 +38,13 @@ class COINS:
...
     edge_gdf : GeoDataFrame
         A GeoDataFrame containing edge geometry of a street network.
+        ``edge_gdf`` cannot contain identical or overlapping LineStrings.
         ``edge_gdf`` should ideally not contain MultiLineStrings.
-    angle_threshold : int, float (default 0)
-        The angle threshold for the COINS algorithm. Segments will only be considered
-        a part of the same street if the deflection angle is above the threshold.
+    angle_threshold : int, float (default 0), units: degrees
+        The threshold for the interior angle within the COINS algorithm.
+        Possible values: `0<=angle_threshold<180`, in degrees.
```
martinfleis, line 45:
```suggestion
        Possible values: ``0 <= angle_threshold < 180``, in degrees.
```

momepy/coins.py
```diff
@@ -378,118 +382,51 @@ def _list_to_pairs(in_list):
...
-        angle = _points_set_angle(line1, line2)
-
-    elif l1p1 == l2p1:
-        if ((l1p1[1] > l1p2[1]) and (l1p1[1] > l2p2[1])) or (
-            (l1p1[1] < l1p2[1]) and (l1p1[1] < l2p2[1])
-        ):
-            angle = 180 - (abs(l1orien) + abs(l2orien))
-        else:
-            angle = _points_set_angle([l1p1, l1p2], [l2p1, l2p2])
+    assert Counter(points.values()) == {1: 2, 2: 1}
```
martinfleis, line 415:
```suggestion
```

momepy/tests/test_coins.py
```diff
@@ -111,4 +112,108 @@ def test_premerge(self):
         )
         assert_index_equal(result.columns, expected_columns)
 
-        assert not result.isna().any().any()
+        assert not result.drop(columns="orientation").isna().any().any()
+
+    def test_sharpangles(self):
```
martinfleis, line 117:
```suggestion
    def test_sharp_angles(self):
```

martinfleis:
One thing I am not super excited about is that this significantly affects performance. On the network from the user guide:

main: `277 ms ± 8.87 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`
this PR: `545 ms ± 12.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`

fnattino:
(sorry for interfering - again)

> One thing I am not super excited about is that this significantly affects performance. On the network from the user guide:

I have noticed a similar performance degradation  when testing for #630 .  I think it is the overhead of creating many very small numpy arrays, as this is carried for each of the line segments in the network. In the test I have run, using native python types and functions from `math` was much faster than bringing the 2-element-long lists to np.ndarray.

martinfleis:
@fnattino thanks for the pointer! With 8e5bf56, the timing goes down to `316 ms ± 1.45 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)` which is acceptable slowdown, I guess.

martinfleis:
And with a6fcacb to `255 ms ± 8.15 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`, so we're actually a little bit faster now :).

jGaboardi:
@fnattino 

> (sorry for interfering - again)

You do not need to apologize and you are not interferring!

martinfleis (approved):
all good with me!

jGaboardi:
How about [this 310 error](https://github.com/pysal/momepy/actions/runs/9700556665/job/26772250213?pr=632#step:5:119)?

martinfleis:
@jGaboardi present on main https://github.com/pysal/momepy/actions/runs/9688801724/job/26747565038 and has nothing to do with our code. xref #634

docs/user_guide/graph/coins.ipynb
jGaboardi, line 1: ## Continuity in street networks
> Momepy allows for the deduction of natural continuity of street networks using the COINS algorithm. The street network is split into individual segments and deflection angles between adjacent segments are computed. Segments are then joined to continuous strokes. Segments will only be considered a part of the same stroke if the deflection angle is above the threshold ***(defaults to zero)***.

Is this still accurate?


## Using ***a*** small network

jGaboardi: I think I answered my own (first) question by reading the docs in `momepy/coins.py`

martinfleis: yeah, it still defaults to zero, we haven't changed that.

jGaboardi: Just one typo I think --> Using ***a*** small network

jGaboardi (approved):

PR #634, opened 2024-06-27
CI: bump minimal requirements as per SPEC 0

martinfleis:
First checking if this actually fixes ci issues.


jGaboardi:
Let's also update [`environment.yml`](https://github.com/pysal/momepy/blob/main/environment.yml)

jGaboardi (approved):

PR #640, opened 2024-07-11
add `mypy` to `pre-commit` - #639

### LINKED ISSUE #639: `mypy` - move from action to `pre-commit`
We should remove the actual `mypy` action that runs and add that to `pre-commit-config`. This can also be configured in `pyproject.toml`

jGaboardi:
This PR resolves #639


martinfleis:
I am not sure I am fan of this... Linting is easy to fix, typing may be harder. And if you have pre-commit hook installed locally, this won't allow you to commit and push to ask in the PR what the typing should be.

PR #642, opened 2024-07-16
REF: refactor remove_false_nodes

martinfleis:
The original implementation tended to leave some nodes when a line had to be merged on both sides. This solves it and as a consequence of refactoring is also about 60x faster.

cc @anastassiavybornova

### DIFF
--- a/momepy/preprocessing.py
+++ b/momepy/preprocessing.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python
 
-import collections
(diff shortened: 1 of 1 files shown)


martinfleis:
Dev failures are not relevant here, I think those are coming from the fix of higher_order in lib.

u3ks (approved):

jGaboardi (approved):

PR #644, opened 2024-07-16
ENH: add flow mode to COINS

martinfleis:
This adds flow mode of continuity which ensures that there's only a one stroke group per LineString and does not cause breakage when dealing with stroke attributes rather than stroke geometries.

### DIFF
--- a/momepy/coins.py
+++ b/momepy/coins.py
@@ -445,7 +445,7 @@ def _angle_between_two_lines(line1, line2):
     dot_product = v1[0] * v2[0] + v1[1] * v2[1]
     norm_v1 = math.sqrt(v1[0] ** 2 + v1[1] ** 2)
     norm_v2 = math.sqrt(v2[0] ** 2 + v2[1] ** 2)
-    cos_theta = dot_product / (norm_v1 * norm_v2)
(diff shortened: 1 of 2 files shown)


momepy/coins.py
```diff
@@ -44,6 +44,16 @@ class COINS:
         Possible values: ``0 <= angle_threshold < 180``, in degrees.
         Segments will only be considered part of the same stroke group
         if the interior angle between them is above the threshold.
+    flow_mode : bool, default False
+        Continuity can be derived based on visiblity or flow. The former creates a
+        stroke group break at any angle above the ``angle_threshold`` no matter if it
+        is on a node where two LineStrings intersect or if it is on an internal node
+        within the LineString. This mimics the "visible continuity". The latter creates
```
jGaboardi, line 51:
```suggestion
        within the LineString. This mimics the "visible continuity." The latter creates
```

martinfleis:
yeah, I wouldn't worry too much about that

jGaboardi (approved):

jGaboardi:
If we don't care about the inability to resolve the windows environment, let's go ahead and merge.

PR #647, opened 2024-07-18
BUG: check loops in remove_false_nodes against any node degree

martinfleis:


martinfleis:
@jGaboardi if you want to ask about tests, I really don't want to do them :D

jGaboardi:
Do we not need a test for this? Or is the test implicit?

jGaboardi:
> @jGaboardi if you want to ask about tests, I really don't want to do them :D

perfect timing -- we commented at exactly the same time

martinfleis:
It is not implicit but given how tests for this are structured now, we would need to alter the test data. Just pretend you missed the missing tests :D

jGaboardi (approved):

PR #648, opened 2024-07-18
TST: Fix expected test values when using higher_order 

u3ks:


martinfleis:
I suppose we can wait for libpysal 4.12 and then pin it. That should happen in a few days.

PR #653, opened 2024-08-26
Allow nodes as an input to graph construction method 'gdf_to_nx'

leigh-pozibl:
Added option to input node features to the method gdf_to_nx

### DIFF
--- a/momepy/utils.py
+++ b/momepy/utils.py
@@ -112,13 +112,13 @@ def _angle(a, b, c):
 
 
 def _generate_primal(
-    graph, gdf_network, fields, multigraph, oneway_column=None, preserve_index=False
+    graph, gdf_edges, gdf_nodes, fields, multigraph, oneway_column=None, preserve_index=False, add_edges_first=True
 ):
     """Generate a primal graph. Helper for ``gdf_to_nx``."""
     graph.graph["approach"] = "primal"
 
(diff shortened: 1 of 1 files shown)


martinfleis:
Hey @leigh-pozibl was this all intentional? Is this something you'd like to see in?

leigh-pozibl:
Hi Martin,

sorry, I inadvertently published that pull request (I'm new to github!)

I have been trying out momepy and wanted to get some data attached to the
nodes in a generated graph. That is useful to me but not sure if that is
something you guys have considered?

apologies for the error.

regards
Leigh

martinfleis:
If you already have nodes and edges, you can create a graph directly using networkx. The purpose of this function is to create a graph from LineString geometry only.

leigh-pozibl:
sure, I liked how your function could build the graph from geometries
though which is my use case.  Is there any particular reason why you don't
consider Point geoms as an input as well? Am I missing something?

For context, I am building graphs of electrical grids, and model the poles
and wires.
regards

martinfleis:
> Is there any particular reason why you don't consider Point geoms as an input as well? Am I missing something?

It is a different use case. The purpose of this function is to facilitate the conversion if you don't have data representing nodes. Because if you do, there's no need for it, you can build the Graph directly with networkx Graph builders.

PR #654, opened 2024-09-12
CI: pin fiona

martinfleis:


martinfleis:
@jGaboardi I'll merge this but note that the same will need to be done across pysal when the oldest envs resolve with pinned geopandas but latest fiona.

PR #658, opened 2024-10-15
ENH: add Streetscape class

martinfleis:
This adds code implemented by Alessandro Araldi and his team, wrapped into a user friendly API that can be accessed from momepy. It essentially enables analysis of streetscape outlined in 10.1111/gean.12416.

The code is sloooooooooooooooooooooooo... ah... oooooooooooooooow. But it runs. Anyone is happy to refactor. I have picked some low hanging fruits but it is in a dire need of complete refactoring for which I do not have a capacity. It should follow the logic we have in `street_profile` and its performance but is now several orders of magnitude slower. But hey, it works and allows reproducible research. so yay!


martinfleis:
@jGaboardi FYI, I am not expecting you to review this.

PR #676, opened 2024-12-10
fix small warning typo in `coins.py`

jGaboardi:
This PR fixes a small warning typo in `coins.py`.


martinfleis:
@jGaboardi You can self-merge stuff like this.

PR #682, opened 2025-02-13
Enhance remove_false_nodes Function: Retain Line Attributes and Expose aggfunc Parameter.

### LINKED ISSUE #447: keep line attributes in `remove_false_nodes`
In [`remove_false_nodes`](https://github.com/pysal/momepy/blob/main/momepy/preprocessing.py#L193), line attributes are lost when merged together. These can be critical (or at least helpful) for further analysis following preprocessing and users should have the option for retaining them, as is done in `osmnx`'s [`simplify_graph`](https://github.com/gboeing/osmnx/blob/9715a9f68dc215c5da6c0e20074647b7569d8dac/osmnx/simplification.py#L268-L279).

speco29:
Fixes #447 

Key Changes:
-Added retain_attributes and aggfunc parameters to the function signature.

-Created a list new_attrs to store the aggregated attributes if retain_attributes is True.

-Aggregated attributes using aggfunc and added them to new_attrs during the loop that processes connected components.

-Combined final_df with the new attributes if retain_attributes is True.

### DIFF
--- a/docs/user_guide/getting_started.ipynb
+++ b/docs/user_guide/getting_started.ipynb
@@ -66,8 +66,7 @@
    "outputs": [],
    "source": [
     "import geopandas as gpd\n",
-    "import momepy\n",
-    "import mapclassify"
+    "import momepy\n"
    ]
   },
   {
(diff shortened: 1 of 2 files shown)


martinfleis:
Hi, sorry. This function has been moved to https://github.com/uscuni/neatnet where the aggregation has been implemented already.

PR #686, opened 2025-02-25
ENH: Added Cellular Automata-based Enclosed Tessellation

yu-ta-sato:
Based on the discussion #684, I added Cellular Automata-based Voronoi diagram for `enclosed_tessellation`, remaining the API the same as before.

Here's a sample code:
```
from momepy import enclosures, enclosed_tessellation
import geopandas as gpd
import matplotlib.pyplot as plt

blg_polygons = [
    shapely.geometry.Polygon([(15, 32), (35, 32), (35, 38), (15, 38)]),
    shapely.geometry.Polygon([(15, 22), (35, 22), (35, 28), (15, 28)]),
    shapely.geometry.Polygon([(15, 92), (35, 92), (35, 98), (15, 98)]),
    shapely.geometry.Polygon([(15, 82), (35, 82), (35, 88), (15, 88)]),
    shapely.geometry.Polygon([(45, 62), (65, 62), (65, 68), (45, 68)]),
    shapely.geometry.Polygon([(45, 52), (65, 52), (65, 58), (45, 58)]),
]
buildings_gdf = GeoDataFrame(
    {
        "building_id": list(range(1, len(blg_polygons) + 1)),
        "geometry": blg_polygons,
    },
    crs="EPSG:3857",
)

# Define sample barrier geometries.
barrier_geoms = [
    shapely.geometry.LineString([(0, 0), (80, 0)]),
    shapely.geometry.LineString([(80, 0), (80, 120)]),
    shapely.geometry.LineString([(80, 120), (0, 120)]),
    shapely.geometry.LineString([(0, 120), (0, 0)]),
    shapely.geometry.LineString([(40, 0), (40, 110)]),
    shapely.geometry.LineString([(10, 30), (40, 30)]),
    shapely.geometry.LineString([(10, 90), (40, 90)]),
    shapely.geometry.LineString([(40, 60), (70, 60)]),
]

barrier_gdf = GeoDataFrame(
    {
        "name": [
            "Bottom Edge",
            "Right Edge",
            "Top Edge",
            "Left Edge",
            "Main Vertical",
            "Left Cul-de-Sac (Bottom)",
            "Left Cul-de-Sac (Top)",
            "Right Cul-de-Sac (Middle)",
        ],
        "geometry": barrier_geoms,
    },
    crs="EPSG:3857",
)
```

There're three patterns of enclosed tessellation with comparison:
```
# Generate enclosure
enclosure = enclosures(barrier_gdf)

# Original enclosed_tessellation function
enclosed_tess_original = enclosed_tessellation(
    buildings_gdf,
    enclosures=enclosure,
)

fig, axes = plt.subplots(1, 3, figsize=(15, 5))

enclosure.plot(ax=axes[0], alpha=0.1)
buildings_gdf.plot(ax=axes[0], color="red")
barrier_gdf.plot(ax=axes[0], color="blue")
enclosed_tess_original.plot(ax=axes[0], alpha=0.3, color="red", edgecolor="black", linewidth=0.5)
axes[0].set_title("Original")

# New enclosed_tessellation function (without inner barriers)
enclosed_tess_no_inner = enclosed_tessellation(
    buildings_gdf,
    enclosures=enclosure,
    use_ca=True,
    cell_size=1.0,
    neighbor_mode="moore",
    fulfill=True)

enclosure.plot(ax=axes[1], alpha=0.1)
buildings_gdf.plot(ax=axes[1], color="red")
barrier_gdf.plot(ax=axes[1], color="blue")
enclosed_tess_no_inner.plot(ax=axes[1], alpha=0.3, color="red", edgecolor="black", linewidth=0.5)
axes[1].set_title("CA, no inner barriers")

# New enclosed_tessellation function (with inner barriers)
enclosed_tess_with_inner = enclosed_tessellation(
    buildings_gdf,
    enclosures=enclosure,
    use_ca=True,
    cell_size=1.0,
    neighbor_mode="moore",
    fulfill=True,
    barriers_for_inner=barrier_gdf)

enclosure.plot(ax=axes[2], alpha=0.1)
buildings_gdf.plot(ax=axes[2], color="red")
barrier_gdf.plot(ax=axes[2], color="blue")
enclosed_tess_with_inner.plot(ax=axes[2], alpha=0.3, color="red", edgecolor="black", linewidth=0.5)
axes[2].set_title("CA, with inner barriers")

plt.tight_layout()
plt.show()
```
output

In the Liverpool City Region by the Overture Maps, the output of the new algorithm with inner barriers (cell size: 1m)are like this:
Screenshot 2025-02-25 at 03 07 40

If it's good to go, let me update the docs.


yu-ta-sato:
@jGaboardi Yes of course! I wanted to be on the same page with you guys in advance, especially the direction of its implementation. If needed, I can separate the function from `enclosed_tessellation`. Will let @martinfleis have a look at the scripts as well.

martinfleis:
Hold on with testing, I'll have some requests but need more time to properly look at it.

martinfleis:
Hey, enclosed tessellation is supposed to be continuous coverage by definition. As illustrated below, that is not the case at the moment. Also, I would very much prefer to ensure that the lines are as straight as possible, without this artifact of the grid we see here. We can either do that in conversion of the grid to polygons or afterwards using coverage simplification. We will need to touch the API as well but let's focus on ensuring this behaves the way we want first.

Screenshot 2025-03-04 at 13 59 35

yu-ta-sato:
> We can either do that in conversion of the grid to polygons or afterwards using coverage simplification.

The cellular-automata grids are already polygonised in the returning outputs (by `_get_cell_polygon` in `_voronoi_by_ca`). For the latter, Ramer–Douglas–Peucker algorithm might be applied for the smoothing?

martinfleis:
> For the latter, Ramer–Douglas–Peucker algorithm might be applied for the smoothing?

No. We need to ensure that whatever comes from the grid is continuous coverage and then use upcoming [coverage_simplify](https://shapely.readthedocs.io/en/latest/reference/shapely.coverage_simplify.html#shapely.coverage_simplify) which ensures that the simplification does not break topological relations between neighbouring cells.

yu-ta-sato:
Did you enable `fulfill=True`? I prepared for the parameter so that the cells of boundaries are assigned to the closest tessellations to ensure the coverage (`_assign_adjacent_seed_cells`). Hope this is what you mean by the continuous coverage

martinfleis:
Ah, no. I did not. However, this should not be an option, it should be set by default.

yu-ta-sato:
Need to chase the on-going discussion on shapely a bit more, but can we implement `coverage_simplify` on the polygons once it gets stable in its v2.1? Meanwhile I can work on the testing and API design.

momepy/functional/_elements.py
```diff
@@ -388,6 +393,7 @@ def _voronoi_by_ca(
     cell_size: float = 1.0,
     neighbor_mode: str = "moore",
     barriers_for_inner: GeoSeries | GeoDataFrame = None,
+    cell_tolerance: float = 2.0,
```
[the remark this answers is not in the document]
yu-ta-sato: Good idea!

momepy/functional/_elements.py
```diff
@@ -332,6 +383,407 @@ def _tess(ix, poly, blg, threshold, shrink, segment, enclosure_id):
...
+        barrier_geoms_buffered = GeoSeries(
+            [poly.buffer(10).exterior for poly in barrier_geoms.geoms],
+            crs=seed_geoms.crs,
+        )
+        barrier_geoms = GeoSeries(barrier_geoms, crs=seed_geoms.crs)
+
+    else:
+        raise ValueError("barrier_geoms must be a Polygon or MultiPolygon")
+
+    outer_union = shapely.ops.unary_union(barrier_geoms_buffered)
```
martinfleis, line 446:
```suggestion
    outer_union = barrier_geoms_buffered.union_all()
```

momepy/functional/_elements.py
```diff
@@ -332,6 +383,407 @@ def _tess(ix, poly, blg, threshold, shrink, segment, enclosure_id):
...
+
+    # Compute inner barriers union if available
+    if (
+        "inner_barriers" in locals()
+        and inner_barriers is not None
+        and not inner_barriers.empty
+    ):
+        inner_union = shapely.ops.unary_union(inner_barriers.geometry)
```
martinfleis, line 454:
```suggestion
        inner_union = inner_barriers.union_all()
```

momepy/functional/_elements.py
```diff
@@ -332,6 +383,407 @@ def _tess(ix, poly, blg, threshold, shrink, segment, enclosure_id):
...
+    else:
+        inner_union = None
+
+    # Combine outer barrier with inner barriers
+    if outer_union and inner_union:
+        prep_barrier = shapely.ops.unary_union([outer_union, inner_union])
```
martinfleis, line 460:
```suggestion
        prep_barrier = shapely.union(outer_union, inner_union)
```

momepy/functional/_elements.py
```diff
@@ -332,6 +383,407 @@ def _tess(ix, poly, blg, threshold, shrink, segment, enclosure_id):
...
+    else:
+        prep_barrier = None
+
+    # Compute grid bounds
+    origin, grid_width, grid_height = _get_grid_bounds(
+        seed_geoms, barrier_geoms_buffered, cell_size
+    )
+
+    # Initialize grid states with UNKNOWN values.
+    states = np.full((grid_height, grid_width), CellState.UNKNOWN.value, dtype=int)
```
martinfleis, line 474: Do we know how small the int can be? To save memory.

momepy/functional/_elements.py
```diff
@@ -332,6 +383,407 @@ def _tess(ix, poly, blg, threshold, shrink, segment, enclosure_id):
...
+        grid_gdf[grid_gdf["site_id"] >= 0]
+        .dissolve(by="site_id")
+        .reset_index()
+        .drop("site_id", axis=1)
+    )
+
+    # Clip by barriers
+    if barrier_geoms is not None and (not barrier_geoms.empty):
+        # Create a union of the barrier geometries.
+        barrier_union = shapely.ops.unary_union(barrier_geoms.geometry)
```
martinfleis, line 570:
```suggestion
        barrier_union = barrier_geoms.union_all()
```

momepy/functional/_elements.py
```diff
@@ -332,6 +383,407 @@ def _tess(ix, poly, blg, threshold, shrink, segment, enclosure_id):
...
+    if SPL_GE_210:
+        # Simplify coverages with coverage_simplify.
+        # torelance set as the square root of the isosceles right triangle with 2 cells_size edges.
+        # For the behavior of coverage_simplify see https://shapely.readthedocs.io/en/latest/reference/shapely.coverage_simplify.html
+        grid_gdf["geometry"] = shapely.coverage_simplify(
+            grid_gdf["geometry"].array, tolerance= ((2 * cell_size) ** 2 / 2) ** 0.5
+        )
+    else:
+        raise ImportError(
+            "Shapely 2.1.0 or higher is required for coverage_simplify."
```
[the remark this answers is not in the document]
martinfleis: Also, the error should say something like `Shapely 2.1.0 or higher is required for tessellation with inner barriers provided.`

momepy/functional/_elements.py
```diff
@@ -332,6 +383,407 @@ def _tess(ix, poly, blg, threshold, shrink, segment, enclosure_id):
...
+        barriers.intersects(enclosure) | barriers.contains(enclosure)
+    ]
+
+    # Clip those segments to stay within the enclosure
+    inner_barriers = gpd.clip(inner_barriers, enclosure)
+
+    # Only keep the geometry which is within the enclosure
+    inner_barriers = inner_barriers[inner_barriers.within(enclosure)]
+
+    return GeoSeries(inner_barriers.geometry, crs=barriers.crs)
```
martinfleis, line 619: Isn't `inner_barriers` already a GeoSeries?

momepy/functional/_elements.py
```diff
@@ -332,6 +383,407 @@ def _tess(ix, poly, blg, threshold, shrink, segment, enclosure_id):
...
+
+
+def _get_cell_polygon(
+    x_idx: int, y_idx: int, cell_size: float = 1.0, origin: Tuple[float, float] = (0, 0)
+) -> shapely.geometry.Polygon:
+    """
+    Generate a grid cell polygon based on the given indices, cell size, and origin.
+    """
+    ox, oy = origin
+    return shapely.geometry.Polygon(
```
martinfleis, line 646:
```suggestion
    return shapely.Polygon(
```

momepy/functional/_elements.py
```diff
@@ -129,6 +134,10 @@ def enclosed_tessellation(
     segment: float = 0.5,
     threshold: float = 0.05,
     n_jobs: int = -1,
+    use_ca: bool = False,
+    cell_size: float = 1.0,
+    neighbor_mode: str = "moore",
+    barriers_for_inner: GeoSeries | GeoDataFrame = None,
```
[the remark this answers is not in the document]
yu-ta-sato: Agree with it. The usage of those parameters (especially `cell_size`) can be explained in docstings and documents.

momepy/functional/_elements.py
```diff
@@ -297,24 +317,55 @@ def enclosed_tessellation(
...
+        if isinstance(poly, shapely.geometry.Polygon):
+            blg = blg[
+                shapely.area(shapely.intersection(blg.geometry.array, poly))
+                > (shapely.area(blg.geometry.array) * threshold)
+            ]
+        else:
+            blg = blg[
+                shapely.area(
+                    shapely.intersection(
+                        blg.geometry.array, shapely.geometry.Polygon(poly.boundary)
```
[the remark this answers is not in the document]
yu-ta-sato: The initial intention was to let the poly be accommodated with inner barriers, but now they're separated. Somehow this part remained, so will remove it.

momepy/functional/_elements.py
```diff
@@ -332,6 +383,407 @@ def _tess(ix, poly, blg, threshold, shrink, segment, enclosure_id):
...
+        # get inner barriers
+        inner_barriers = _get_inner_barriers(
+            barrier_geoms, barriers_for_inner.to_crs(seed_geoms.crs)
+        )
+
+    # Handle barrier_geoms if it is a Polygon or MultiPolygon
+    if barrier_geoms.geom_type == "Polygon":
+        # Take buffer of polygon and extract its exterior boundary
+        barrier_geoms_buffered = GeoSeries(
+            [barrier_geoms.buffer(10).exterior], crs=seed_geoms.crs
```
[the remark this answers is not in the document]
yu-ta-sato: This amount of buffer was needed to ensure that all the extent are assigned to either of the cell state, which could be arbitrary number (This kind of vacant cells occurred between three tessellations). Will set the number based on the cell size automatically.

momepy/functional/_elements.py
yu-ta-sato: It's four types defined for CellState, which is used before the assignment. Once a cell is assigned to any of sites, a non-negative integer (id of the building) is used in the following processes. It's for readability, so you can consider it as merely an equivalent to negative integers.

yu-ta-sato: The point is, it was hard to read and recognise them from integers (-1: unknown, -2: boundary, -3: barrier, -4: frontier)

momepy/functional/_elements.py
yu-ta-sato: I checked it, you're right.

martinfleis:
Hey, I pushed some changes and got this up to speed with main. 

The main issue I see now is performance. On the tiny bubenec dataset, this algorithm takes `13.4 s ± 1.23 s per loop (mean ± std. dev. of 7 runs, 1 loop each)` while the standard one `172 ms ± 57.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`. That is two orders of magnitude slower. @jGaboardi how do you feel about that?

momepy/elements.py
```diff
@@ -206,6 +212,17 @@ def enclosed_tessellation(
...
+        Barriers that should be included in the tessellation process. By passing
+        inner barriers, tessellation will be derived using less performant
+        Cellular Automata implememtation but it will recognise dangling barries as valid
+        limits of the cell growth. See the user guide for details. By default None.
+    cell_size : float, optional
+        Grid cell size when ``inner_barriers`` is not None. Otherwise ignored.
+        By default 1.0
+    neighbor_mode : str, optional
+        Choice of neighbor connectivity ('moore' or 'neumann') when ``inner_barriers``
+        is not None. Otherwise ignored.. By default 'moore'.
```
jGaboardi, line 225:
```suggestion
        is not None. Otherwise ignored. By default 'moore'.
```

jGaboardi:
> That is two orders of magnitude slower. @jGaboardi how do you feel about that?

I am not excited about it, but since it is optional functionality maybe OK?

martinfleis:
Good job!

@yu-ta-sato can you also try to craft some tests for this and expand the user guide on enclosed tessellation? You can add the test to the existing block.

yu-ta-sato:
Somehow the current code penetrates through the inner barriers from the connected point with the outers.
I'm still investigating the root cause, but it seems it was already happening even after reverting back to before I vectorised the polygon generation.

image

yu-ta-sato:
Tests and docstrings are done. Tests are not covered for line 606 and 619 in `elements.py` from the test cases prepared.

martinfleis (approved):
Looks good to me now. @yu-ta-sato let us know if there's something you still want to do here.

martinfleis:
That is the plan. This has been hanging around for too long.

jGaboardi:
@yu-ta-sato -- Is this is in order to be merged?

@martinfleis -- Shall we get this merged sooner than later?

jGaboardi (approved):

yu-ta-sato:
@martinfleis @jGaboardi, it's my sincere apologies for the delayed check, and thank you so much for following up this pull request! I check the updates and confirmed that the shapely (GEOS) issue is solved with linting & typing. Really appreciate all the support!!

PR #696, opened 2025-04-17
DEP: deprecate functions moved to neatnet

martinfleis:
@jGaboardi what are we doing with `roundabout_simplification` and `consolidate_intersections`? If we want to keep them alive, then both should likely move to neatnet.

### DIFF
--- a/momepy/preprocessing.py
+++ b/momepy/preprocessing.py
@@ -180,6 +180,14 @@ def remove_false_nodes(gdf):
     momepy.extend_lines
     momepy.close_gaps
     """
+    warnings.warn(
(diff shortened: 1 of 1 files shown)


jGaboardi:
> @jGaboardi what are we doing with `roundabout_simplification` and `consolidate_intersections`? If we want to keep them alive, then both should likely move to neatnet.

Good question...

I think it's your call on that. What are you leaning towards?

martinfleis:
Never used them in any pipeline but I can imagine ppl might want to do just a subset of simplification. I am also not sure how well they work. We can postpone this decision for a next release if needed. There's no rush to remove stuff.

PR #698, opened 2025-04-22
MAINT: add more pytest.approx calls

cjwatson:
These tests failed on Debian i386, probably due to x87 excess precision: https://bugs.debian.org/1103140


martinfleis:
@cjwatson do you need a release?

cjwatson:
I think we can just cherry-pick that patch.  No urgency about a release unless you were planning on doing one anyway.

PR #706, opened 2025-08-12
DEP: remove deprecated class API

martinfleis:
Boom. All gone.


jGaboardi:
Quite the heroic effort @martinfleis !

PR #707, opened 2025-08-13
DEP: fix user guide

martinfleis:
Follow up on #706


docs/user_guide/preprocessing/roundabout_simplification.ipynb
jGaboardi, line 1: @martinfleis Are you OK having the `TqdmWarning` after imports displayed in the notebook that points back to your local directory?

PR #712, opened 2025-10-05
Morphotope tutorial

u3ks:
The notebook has:
- same place as the clustering tutorial and same variables
- Overture data
- Neatnet
- ETC + adaptive limit + postprocessing
- No spatial lag
- S3
- Dealing with noise

Anything missing ? maybe itd be good to cache the overture data, because it takes 4-5 minutes just reading it ...


martinfleis:
Can we do OSM data and osmnx as we do elsewhere in the guide? Any reason to use Overture?

martinfleis:
That is fine, we don't need to be 1:1 with the paper here. The other is not either.

Issue #718, opened 2025-10-10
Artefacts in output when tessellating islands with narrow channels


jontwo:
### Describe the problem

Running `morphological_tessellation` on some island coastlines, there were some strange artefacts in the output where the channel was <100m wide.

Input island coastlines.

<img width="658" height="459" alt="Image" src="https://github.com/user-attachments/assets/59d1f55e-689d-4cc4-beca-585f4a79faa0" />

Tessellation output overlaid on the coastlines, the narrow point is around 60m wide.

<img width="658" height="459" alt="Image" src="https://github.com/user-attachments/assets/8e844c0f-95d0-4b12-91df-af890d089dd5" />

### Steps to reproduce

```python

import geopandas
import momepy

coastline = "path/to/coastline.shp"
polygons = "path/to/output.shp"

coast_gdf = geopandas.read_file(coastline).reset_index(drop=True)
result = momepy.morphological_tessellation(coast_gdf, segment=500, shrink=20)
result.to_file(polygons)
```

### Versions of your packages

```
python=3.12.11
momepy=0.9.1
geopandas=1.1.1
libpysal=4.13.0
networkx=3.5
```

### Your operating system

Ubuntu 22.04.5 LTS

### Additional context

Having reread the [doc](https://docs.momepy.org/en/stable/api/momepy.morphological_tessellation.html), I can see the values we were using for `segment` and `shrink` are way off the default values. They had been working for us up to that point. When I lowered the `segment` value to 100, the issue went away.

So, you can probably close this issue as being user error, but could I ask for a clarification in the docs? Both `segment` and `shrink` represent a distance, but is this map units? I'm assuming it is, but it would be good to clarify it, as it could be pixels or something else.

PR #722, opened 2025-11-07
DOC: Add missing text to enclosed_tessellation

jontwo:
Continuing #718, this PR adds in some missing text to the docstring.

### DIFF
--- a/momepy/elements.py
+++ b/momepy/elements.py
@@ -78,8 +78,7 @@ def morphological_tessellation(
         a sweet spot between computational inefficiency (when the value is too low)
         and suboptimal resulting geometry (when the value is too large). The default
         is empirically derived for the use case on building footprints represented in
-        map units.
-        By default 0.5
+        map units. By default 0.5
     simplify: bool, optional
         Whether to attempt to simplify the resulting tesselation boundaries with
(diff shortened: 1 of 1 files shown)


momepy/elements.py
```diff
@@ -201,7 +201,8 @@ def enclosed_tessellation(
         The maximum distance between points after discretization. A right value is
         a sweet spot between computational inefficiency (when the value is too low)
         and suboptimal resulting geometry (when the value is too large). The default
-        is empirically derived for the use case on
+        is empirically derived for the use case on building footprints represented in
+        meters. By default 0.5
```
martinfleis: Not necessarily. Could be feet or something. It is represented in units of the CRS.

jGaboardi (changes requested):
Thanks of this @jontwo !

momepy/elements.py
jontwo: Sure, I was copying from the docstring on line 80, which was essentially the same. I can update both to map units.

jGaboardi:
@jontwo -- I think these edits are nice. Thank you for your attention to this. Let's put `By default 0.5` directly after `map units. ` on the same line with no break.

jGaboardi (approved):

PR #726, opened 2025-12-23
COMPAT: streetscape compatibility with numpy 2.4

martinfleis:
Resolves the failures and warnings coming from CI and the notebook.


jGaboardi (approved):
🙇

PR #753, opened 2026-04-25
Reup spec000 [2026-04-25]

### LINKED ISSUE #752: simplify by default in `*_tessellation()`
@martinfleis -- I am working on re-upping spec000, which would bring `shapely==v2.1` up to the oldest version.

I am seeing these notes in [`morphological_tessellation()`](https://github.com/pysal/momepy/blob/4f9def7f159e39add8f94a95c69d4dbbbf0225f1/momepy/elements.py#L120-L126) and [`enclosed_tessellation()`](https://github.com/pysal/momepy/blob/4f9def7f159e39add8f94a95c69d4dbbbf0225f1/momepy/elements.py#L267-L273).

```
# TODO: remove the keyword and do simplification by default once it is
# safe to pin shapely 2.1
```

So we want to do this, correct?:

1. Give a deprecation warning that the `simplify` removed in a later release and hardcode `simplify=True` within both the functions
2. Remove the key word entirely, maybe after the next release

jGaboardi:
* Reup spec000 [2026-04-25]
* standardized/ sync envs
* resolves #752


momepy/elements.py
```diff
@@ -409,6 +413,9 @@ def _tess(
         Threshold for building inclusion.
     shrink : float
         Shrink distance for tessellation.
+    to_simplify : bool
+        Flag for simplification
+        **NOW ONLY SIMPLIFYING** Remove option entirely here?
     segment : float
```
jGaboardi, line 423: @martinfleis - should we just go ahead and remove the `to_simplify` positional argument from `_tess()` now?

jGaboardi:
re: [312-oldest failure](https://github.com/pysal/momepy/actions/runs/24941601902/job/73036136599#step:5:534)

* is this perhaps due the [referenced GEOS bug](https://github.com/pysal/momepy/blob/c988ecedb057306dc3eb9989d78734a1cf65c95b/momepy/tests/test_elements.py#L199)
* or due to something else entirely?

```
In [1]: import shapely

In [2]: shapely.geos_version
Out[2]: (3, 14, 1)
```

```
geometry = 0    POLYGON ((30.415 30, 29.923 30, 29.431 30, 28....
1    POLYGON ((15.646 30, 16.138 30, 16.631 30, 17....
2    POLYGON ((49.64 94.055, 49.32 93.582, 49.011 9...
3    GEOMETRYCOLLECTION (POLYGON ((15.646 90, 16.13...
4    POLYGON ((30.989 63.109, 30.68 63.582, 30.36 6...
5    POLYGON ((47.123 39.704, 46.631 40.112, 46.138...
6    POLYGON ((15.646 64.851, 16.138 64.989, 16.631...
7    POLYGON ((64.354 77.326, 63.862 76.964, 63.369...
```

martinfleis:
I have no idea given the geos and shapely in the oldest is the same as in the latest.

martinfleis:
I have wrapped the simplification in try/except as it can be brittle. The remaining issue in oldest is of unknown origin... I'd just skip the geometry type check.

momepy/tests/test_elements.py
```diff
@@ -289,7 +290,9 @@ def test_enclosed_tessellation_inner_barrier_cellular(self):
             threshold=None,
             n_jobs=1,
         )
-        assert set(tess_empty.geom_type.unique()) <= {"Polygon", "MultiPolygon"}
+        if SHP_GE_2_2:
```
martinfleis: I don't think this is a correct condition but I am not sure what is. It is not tied to Shapely/GEOS as our due to the slow release pace of Shapely, our oldest env uses the same as our latest. It is Scipy or something else causing this and I can't be bothered figuring out what. I know it is not shapely.

jGaboardi: Agreed. That was simply "a condition" that was at least tangentially tied to the failure. Not optimal, but would probably take more time to figure out than it's worth.

momepy/tests/test_elements.py
```diff
@@ -287,14 +284,15 @@ def test_enclosed_tessellation_inner_barrier_cellular(self):
...
             buildings,
             enclosures,
-            simplify=False,
             inner_barriers=empty_barriers,
             cell_size=1,
             neighbor_mode="moore",
             threshold=None,
             n_jobs=1,
         )
-        assert set(tess_empty.geom_type.unique()) <= {"Polygon", "MultiPolygon"}
```
martinfleis, line 297:
```suggestion
        if not set(tess_empty.geom_type.unique()) <= {"Polygon", "MultiPolygon"}:
            pytest.xfail("Tessellation produced GeometryCollection for an unknown reason")
```

jGaboardi: Screenshot 2026-04-27 at 12 16 12 PM

What the hell is up with GH recently... Second time this has happened in a month our so...

jGaboardi: applied suggestion locally & remove conditional

PR #754, opened 2026-05-11
CI: remove iprogress

martinfleis:
Trying to figure out what was this needed for. It was last updated in 2016 and is causing trouble in env resolution.


martinfleis:
Seems all good, no warning or nothing.

PR #756, opened 2026-05-21
REF: simplify implementation of enclosures

martinfleis:
I hit a minor issue when passing polygons instead of linestrings as additional barriers and while looking for a cause, I realised that this can be significantly simplified, while ensuring that polygons are treated as a linework.

It is also faster, from 45.7 ms ± 2.6 ms  to  17.8 ms ± 112 μs.


jGaboardi (approved):

jGaboardi:
Nice!

PR #758, opened 2026-06-02
REF: refactor Streetscape to increase its performance

martinfleis:
I wanted to include `Streetscape` in some teaching material I am currently working on but I hit its super slow performance. I knew it is slow and I knew the code was, say, suboptimal. So, I pointed Codex at it to see how far can we go. I did not understand the old code and I think I understand the new one a bit more. What is important, is that on the Kahla data from the docs, we get this:

main: `182.27s`
current: `3.47s`

And with the small bubenec data, which also includes `compute_plots` and `compute_slope` bits, we get:

main: `2.420s`
current: `0.299s`

So while I am not super excited that this is Codex-written, it is clearly beneficial here.

For a context - we were given old spaghetti code from a research paper introducing the streetscape metrics. I had my undergrad student to translate that from a bunch of undocumented notebooks into something reusable. That is roughly what is on main. And it is dead slow. I couldn't force myself to refactor it so I forced Codex to do it. And that is a what is in the PR. It is completely generated but performance gains are significant and the code is actually a bit more understandable than it was before (which says more about the state of the original than of qualities of Codex). I just want to make sure we're fine with this approach, for these unique edge cases.

@jGaboardi thoughts?

### DIFF
--- a/momepy/streetscape.py
+++ b/momepy/streetscape.py
@@ -243,9 +243,6 @@ class Streetscape:
         # semi_ortho_segment_size = self.sightline_spacing/2
         semi_ortho_segment_size = self.intersection_offset / 2
 
-        field_geometry = 0
-        field_uid = 1
-
         seg_st = shapely.line_interpolate_point(
             line, distances - semi_ortho_segment_size
         )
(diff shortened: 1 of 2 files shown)


jGaboardi:
I think acceptable in case like these, especially since you declare that you used Codex and Codex itself did not make the PR.

This is an insane speed up.

jGaboardi (approved):

jGaboardi (approved):

